PR #15, opened 2020-03-12
Add support for auth

jhkldev:


app/config/AppConfig.scala
```diff
@@ -29,4 +29,6 @@ class AppConfig @Inject()(config: Configuration, servicesConfig: ServicesConfig)
   val graphiteHost: String     = config.get[String]("microservice.metrics.graphite.host")
 
   val traderAtDestinationUrl   = servicesConfig.baseUrl("transit-movement-trader-at-destination")
+
+  lazy val enrolmentKey: String     = config.get[String]("urls.enrolmentKey")
```
TheNickWilson: Could we put this in a better place in config, as it's not a URL?

app/controllers/actions/AuthAction.scala
```diff
@@ -0,0 +1,63 @@
...
+import play.api.mvc.Results._
+import play.api.mvc._
+import uk.gov.hmrc.auth.core._
+import uk.gov.hmrc.auth.core.retrieve.v2.Retrievals
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.HeaderCarrierConverter
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait IdentifierAction extends ActionBuilder[IdentifierRequest, AnyContent] with ActionFunction[Request, IdentifierRequest]
```
TheNickWilson: I don't think we need this trait; we're directly using the `AuthAction` in our controller.  We could just have the `AuthAction` extend `ActionBuilder[IdentifierRequest, AnyContent] with ActionFunction[Request, IdentifierRequest]` instead.

app/controllers/actions/AuthAction.scala
```diff
@@ -0,0 +1,63 @@
...
+    implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromHeadersAndSession(request.headers, Some(request.session))
+
+    authorised(Enrolment(config.enrolmentKey)).retrieve(Retrievals.authorisedEnrolments) {
+      enrolments: Enrolments =>
+        val eoriNumber: String = (for {
+          enrolment  <- enrolments.enrolments.find(_.key.equals(config.enrolmentKey))
+          identifier <- enrolment.getIdentifier(enrolmentIdentifierKey)
+        } yield identifier.value).getOrElse(throw InsufficientEnrolments(s"Unable to retrieve enrolment for $enrolmentIdentifierKey"))
+
+        block(IdentifierRequest(request, Some(eoriNumber)))
```
TheNickWilson: We don't need `IdentifierRequest` as we already have `AuthRequest`; could we swap that out please (and change the signature of the `block` on line 44 as well)?  `AuthRequest` has the eoriNumber as a `String` rather than `Option[String]` so that'll need to be tweaked too.

app/models/requests/IdentifierRequest.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.requests
+
+import play.api.mvc.{Request, WrappedRequest}
+
+case class IdentifierRequest[A](request: Request[A], eoriNumber: Option[String]) extends WrappedRequest[A](request)
```
TheNickWilson: We don't need this class

conf/application.conf
```diff
@@ -186,3 +188,7 @@ microservice {
     }
   }
 }
+
+urls {
+  enrolmentKey  = "HMCE-NCTS-ORG"
```
TheNickWilson: Could we move this outside of the `urls` node?

test/controllers/ArrivalsControllerSpec.scala
```diff
@@ -16,27 +16,52 @@
...
+  val cc = stubControllerComponents()
+
+  val config: AppConfig = app.injector.instanceOf[AppConfig]
+
+  val parser = app.injector.instanceOf[BodyParsers.Default]
+
+  val authConnector = mock[AuthConnector]
+  val fakeAuthAction = FakeAuthAction(authConnector, config, parser)
+
+  object TestArrivalsController extends ArrivalsController(cc, fakeAuthAction, xmlValidation)
```
TheNickWilson: Rather than creating a test object that extends the controller we want to test, could we test the controller itself?

We should be able to remove lines 40 to 56, and instead have something that overrides `AuthAction` with our `FakeAuthAction` in the app.  This will be something like:

```
  override lazy val app = GuiceApplicationBuilder()
    .overrides(bind[AuthAction].to[FakeAuthAction])
    .build()
```

This will mean that whenever our tests run, they're using the `FakeAuthAction` instead of the real one, and we avoid calling off to auth etc.  There'll be a few more changes to the spec as well which are commented below

test/controllers/ArrivalsControllerSpec.scala
```diff
@@ -16,27 +16,52 @@
...
+
+  def ctcFakeRequest() = FakeRequest(method = "POST", uri = "", headers = FakeHeaders(), body = AnyContentAsEmpty)
```
TheNickWilson: All of the `uri`'s (there are a few other examples in this file) should be the real ones rather than empty strings - i.e. you can use `uri = routes.ArrivalsController.createArrivalNotification().url`

test/controllers/ArrivalsControllerSpec.scala
```diff
@@ -76,36 +101,79 @@ class ArrivalsControllerSpec extends FreeSpec with MustMatchers with GuiceOneApp
          </CUSOFFPREOFFRES>
        </CC007A>
      )
-     val result = route(app, request).value
+     val result = TestArrivalsController.createArrivalNotification().apply(request)
```
TheNickWilson: In each of the tests, we should replace this line with `val result = route(app, request).value`

test/controllers/actions/FakeAuthAction.scala
```diff
@@ -0,0 +1,37 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+case class FakeAuthAction @Inject()(
+                                override val authConnector: AuthConnector,
+                                config: AppConfig,
+                                override val parser: BodyParsers.Default
+                          )(implicit override val executionContext: ExecutionContext)
+  extends AuthAction(authConnector, config, parser) {
+
+  override def invokeBlock[A](request: Request[A], block: IdentifierRequest[A] => Future[Result]): Future[Result] = {
```
TheNickWilson: This should return an `AuthAction` rather than an `IdentifierAction`

jhkldev:
I've made the requested changes.

app/controllers/actions/AuthRequest.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.actions
+
+import play.api.mvc.{Request, WrappedRequest}
+
+case class AuthRequest[A](request: Request[A], eori: Option[String]) extends WrappedRequest[A](request)
```
TheNickWilson: `eori` here could be a String, not an Option[String]

PR #19, opened 2020-03-16
Return Location header on successful IE007 POST

jhkldev:


app/controllers/ArrivalsController.scala
```diff
@@ -16,37 +16,54 @@
...
 import scala.concurrent.{ExecutionContext, Future}
 import scala.xml.NodeSeq
 
-
 class ArrivalsController @Inject()(cc: ControllerComponents,
                                    authAction: AuthAction,
                                    messageConnector: MessageConnector,
                                    xmlValidationService: XmlValidationService)(implicit ec: ExecutionContext) extends BackendController(cc) {
+
+  private val arrivalsFormat = """^/arrivals/(.*)$""".r
```
crystalmark: Could be that the Location received from the BE service has parameters which we don't want to pass on to the 3rd party.  e.g. /movements/arrivals/123?status=success
That _should_ never happen but lets program defensively and terminate the arrival id at a ? character

PR #38, opened 2020-04-21
Change downstream response code checks

jhkldev:


app/connectors/ArrivalConnector.scala
```diff
@@ -38,6 +38,11 @@ class ArrivalConnector @Inject()(http: HttpClient, appConfig: AppConfig) {
     http.POSTString(url, message)(HttpReads.readRaw, implicitly, implicitly)
   }
 
+  def postToMessagesRoute(message: String, arrivalId: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] = {
```
crystalmark, line 41: Since calls to downstream /messages is going to only get bigger, it may be worth considering an ArrivalMessageController

app/controllers/ArrivalsController.scala
```diff
@@ -61,8 +59,12 @@ class ArrivalsController @Inject()(cc: ControllerComponents,
             case e: Upstream4xxResponse =>
```
crystalmark, line 59: Could these case+if+else if be refactored into for comprehensions?

app/connectors/ArrivalConnector.scala
jhkldev: :thumbsup:

The code could be refactored to have ArrivalMovementController and ArrivalMessageController

app/controllers/ArrivalsController.scala
jhkldev: There are plans to make 4xx and 5xx responses use status codes. They're currently exceptions. When this change occurs, it should improve the handling of downstream responses.

PR #40, opened 2020-04-23
Refactor to enable more message types to be supoorted in POST

jhkldev:


app/controllers/ArrivalMessagesController.scala
```diff
@@ -39,28 +39,33 @@ class ArrivalMessagesController @Inject()(cc: ControllerComponents,
...
-                  case Some(locationValue) => Utils.arrivalId(locationValue, fragmentIndex = -2) match {
-                    case Success(id) =>
-                      Accepted.withHeaders(LOCATION -> s"/customs/transits/movements/arrivals/${Utils.urlEncode(id)}/messages")
-                    case Failure(_) =>
-                      InternalServerError
-                  }
-                  case _ =>
-                    InternalServerError
+      request.body.head.label match {
+        case "CC044A" =>
```
RobertLaverick, line 45: I'm really not a fan of this, partially for being a magic string (it'd be nice to have it form part of the XSD type as a field) but it'd be nicer to be able to just match against the parse result like you showed me `XSDSuccessful(type) match` or something like that.

app/controllers/ArrivalMessagesController.scala
```diff
@@ -39,28 +39,33 @@ class ArrivalMessagesController @Inject()(cc: ControllerComponents,
...
-              case _ => Status(response.status)
-            }
+              }
+            case Left(_) =>
+              Future.successful(BadRequest)
           }
-        case Left(_) =>
-          Future.successful(BadRequest)
+        case _ =>
+          Future.successful(NotImplemented)
```
RobertLaverick, line 68: I get why you've done this - but it might make more sense to do "NotSupported" instead. I think that's a response code.

crystalmark: NotSupported implies that we're not going to do it.  NoImplemented means to me Not Implemented **Yet**

RobertLaverick: Fair enough, stick with it.

app/controllers/actions/XSDValidationActions.scala
```diff
@@ -0,0 +1,81 @@
...
+        Future.successful(Left(BadRequest))
+    }
+  }
+}
+
+class ValidateMessage @Inject()(xmlValidationService: XmlValidationService)(
+  implicit val executionContext: ExecutionContext)
+  extends ActionRefiner[Request, Request] {
+
+  private val supportedMessageTypes = Map("CC044A" -> UnloadingRemarksXSD)
```
RobertLaverick, line 55: You could build this on an XSDFile object if you add a "tag" or "label" field to the XSD trait as well, and then you only need to call XSDFile.tagToTypeMap (or something like that)

RobertLaverick:
:shipit:

PR #45, opened 2020-04-30
Specify correct content type in MessageConnector

jhkldev:
MessageConnector wasn't sending the correct content type in the header so downstream was responding with 415 Unsupported Media Type.<br><br>The CC044A example in TestXml has also been updated to contain the correct date field as downstream was failing with BadRequest without it.

### DIFF
--- a/app/connectors/MessageConnector.scala
+++ b/app/connectors/MessageConnector.scala
@@ -27,13 +27,13 @@ import uk.gov.hmrc.play.bootstrap.http.HttpClient
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class MessageConnector @Inject()(http: HttpClient, appConfig: AppConfig) extends HttpErrorFunctions {
+class MessageConnector @Inject()(http: HttpClient, appConfig: AppConfig) extends BaseConnector with HttpErrorFunctions {
 
   val rootUrl = appConfig.traderAtDestinationUrl + "/transit-movements-trader-at-destination/movements"
 
   def get(arrivalId: String, messageId: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[HttpResponse, MovementMessage]] = {
     val url = rootUrl + s"/arrivals/$arrivalId/messages/$messageId"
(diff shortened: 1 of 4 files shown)


app/connectors/MessageConnector.scala
```diff
@@ -32,7 +33,7 @@ class MessageConnector @Inject()(http: HttpClient, appConfig: AppConfig) extends
 
   def get(arrivalId: String, messageId: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[HttpResponse, MovementMessage]] = {
     val url = rootUrl + s"/arrivals/$arrivalId/messages/$messageId"
-    http.GET[HttpResponse](url)(CustomHttpReader, implicitly, implicitly).map { r =>
+    http.GET[HttpResponse](url, queryParams = Seq(), headers = Seq(HeaderNames.CONTENT_TYPE -> "application/json"))(CustomHttpReader, implicitly, implicitly).map { r =>
```
RobertLaverick, line 36: If we're going to do this in all connectors, it might be worth having a base connector that can define a def for addHeaders that can take in the content type as a MimeType (see the ArrivalConnector for the type). That should tidy up some of this and remove the magic strings

test/data/TestXml.scala
```diff
@@ -89,7 +89,7 @@ trait TestXml {
       <MesRecMES6>111111</MesRecMES6>
       <!--Optional:-->
       <RecIdeCodQuaMES7>1111</RecIdeCodQuaMES7>
-      <DatOfPreMES9>111111</DatOfPreMES9>
+      <DatOfPreMES9>20001001</DatOfPreMES9>
```
RobertLaverick, line 92: ?

app/connectors/MessageConnector.scala
jhkldev: For now can we add headers as a private val outside the methods and reference this in the methods?

RobertLaverick: I'm confused, that's practically the same quantity of work.

jhkldev: OK. So we could introduce a BaseConnector trait with a headers val?

jhkldev: Added BaseConnector trait

PR #50, opened 2020-05-07
Integrated with backend branch for GET arrival messages endpoint

RobertLaverick:


app/connectors/MessageConnector.scala
```diff
@@ -47,4 +47,17 @@ class MessageConnector @Inject()(http: HttpClient, appConfig: AppConfig) extends
...
+    val url = rootUrl + s"/arrivals/${Utils.urlEncode(arrivalId)}/messages"
+    http.GET[HttpResponse](url, queryParams = Seq(), responseHeaders())(CustomHttpReader, implicitly, implicitly).map { r =>
+      if(is2xx(r.status)) {
+        r.json.asOpt[Arrival] match {
+          case Some(x) => Right(x)
+          case _ => Left(CustomHttpReader.recode(INTERNAL_SERVER_ERROR, r))
+        }
+      } else Left(r)
+    }
+  }
```
jhkldev, line 61: I think the short named variables in the PR should be changed to the full name representing the type.

app/controllers/ArrivalMessagesController.scala
```diff
@@ -68,4 +68,32 @@ class ArrivalMessagesController @Inject()(cc: ControllerComponents,
...
+                    Utils.lastFragment(m.location) match {
+                      case Failure(_) => throw new Exception("parsing location header failed")
+                      case Success(value) => ResponseMessage(m) copy (location = s"/movements/arrivals/${Utils.urlEncode(arrivalId)}/messages/$value")
+                    }
+                  }
+                  Ok(Json.toJson(ResponseArrival(a).copy(arrival = s"/movements/arrivals/${Utils.urlEncode(arrivalId)}", messages = messages)))
+                }
+                catch {
+                  case _ => InternalServerError
+                }
```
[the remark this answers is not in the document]
RobertLaverick: Specifically, it can't - because it is a map over a sequence, rather than a future.

app/connectors/MessageConnector.scala
RobertLaverick: I'm following the pattern that I followed previously in this file. I'll change those as well.

RobertLaverick: Actually, I think I have something I can deduplicate here, so I might try that.

app/controllers/ArrivalMessagesController.scala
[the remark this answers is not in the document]
jhkldev: We can add a lastFragmentSafe method if we know the location from downstream will always be valid and then we can remove the try/catch block completely.

```diff
diff --git a/app/controllers/ArrivalMessagesController.scala b/app/controllers/ArrivalMessagesController.scala
index cc041da..20dbe78 100644
--- a/app/controllers/ArrivalMessagesController.scala
+++ b/app/controllers/ArrivalMessagesController.scala
@@ -19,6 +19,7 @@ package controllers
 import connectors.MessageConnector
 import controllers.actions.{AuthAction, ValidateAcceptJsonHeaderAction, ValidateMessageAction}
 import javax.inject.Inject
+import models.domain.Arrival
 import models.response.{ResponseArrival, ResponseMessage}
 import play.api.libs.json.Json
 import play.api.mvc.{Action, AnyContent, ControllerComponents, Result}
@@ -72,28 +73,20 @@ class ArrivalMessagesController @Inject()(cc: ControllerComponents,
   def getArrivalMessages(arrivalId: String): Action[AnyContent] =
     (authAction andThen validateAcceptJsonHeaderAction).async {
       implicit request => {
-          messageConnector.getArrivalMessages(arrivalId).map { r =>
-            r match {
-              case Right(a) => {
-
-                try {
-                  val messages = a.messages.map { m =>
-                    Utils.lastFragment(m.location) match {
-                      case Failure(_) => throw new Exception("parsing location header failed")
-                      case Success(value) => ResponseMessage(m) copy (location = s"/movements/arrivals/${Utils.urlEncode(arrivalId)}/messages/$value")
-                    }
+          messageConnector.getArrivalMessages(arrivalId).map {
+            response =>
+              response match {
+                case Left(response) =>
+                  handleNon2xx(response)
+                case Right(arrival) => {
+                  val messages = arrival.messages.map {
+                    m =>
+                      ResponseMessage(m) copy (location = s"/movements/arrivals/${Utils.urlEncode(arrivalId)}/messages/${Utils.lastFragmentSafe(m.location)}")
                   }
-                  Ok(Json.toJson(ResponseArrival(a).copy(arrival = s"/movements/arrivals/${Utils.urlEncode(arrivalId)}", messages = messages)))
-                }
-                catch {
-                  case _ => InternalServerError
+                  Ok(Json.toJson(ResponseArrival(arrival).copy(arrival = s"/movements/arrivals/${Utils.urlEncode(arrivalId)}", messages = messages)))
                 }
-
               }
-              case Left(response) => handleNon2xx(response)
-
           }
-        }
       }
     }
 }
diff --git a/app/utils/Utils.scala b/app/utils/Utils.scala
index addda45..850107a 100644
--- a/app/utils/Utils.scala
+++ b/app/utils/Utils.scala
@@ -25,7 +25,10 @@ object Utils {
 
   def lastFragment(location: String): Try[String] =
     Try(new URI(location).getPath.split("/").last)
-  
+
+  def lastFragmentSafe(location: String): String =
+    new URI(location).getPath.split("/").last
+
   def urlEncode(str: String): String =
     URLEncoder.encode(str, "UTF-8")
 }
diff --git a/test/controllers/ArrivalMessagesControllerSpec.scala b/test/controllers/ArrivalMessagesControllerSpec.scala
index b9b51dc..cc5085d 100644
--- a/test/controllers/ArrivalMessagesControllerSpec.scala
+++ b/test/controllers/ArrivalMessagesControllerSpec.scala
@@ -254,15 +254,5 @@ class ArrivalMessagesControllerSpec extends FreeSpec with MustMatchers with Guic
 
       status(result) mustBe INTERNAL_SERVER_ERROR
     }
-
-    "return 500 if downstream provides an unsafe message header" in {
-      when(mockMessageConnector.getArrivalMessages(any())(any(), any()))
-        .thenReturn(Future.successful(Right(sourceArrival.copy(messages = Seq(sourceMovement.copy(location = "/transit-movements-trader-at-destination/movements/arrivals/<>"))))))
-
-      val request = FakeRequest("GET", "/movements/arrivals/123/messages", headers = FakeHeaders(Seq(HeaderNames.ACCEPT -> "application/vnd.hmrc.1.0+json")), AnyContentAsEmpty)
-      val result = route(app, request).value
-
-      status(result) mustBe INTERNAL_SERVER_ERROR
-    }
   }
 }

```

PR #57, opened 2020-05-15
Don't modify HeaderCarrier when sending headers

jhkldev:


RobertLaverick:
I'm ok with approving this, though I'm confused why this is the solution if it works locally

PR #73, opened 2020-05-21
CTDA-183: Return 406 Not Acceptable when invalid Accept header has be…

jhkldev:
…en specified

### DIFF
--- a/app/utils/Utils.scala
+++ b/app/utils/Utils.scala
@@ -23,6 +23,8 @@ import scala.util.Try
 object Utils {
   val acceptHeaderPattern = "^application/vnd[.]{1}hmrc[.]{1}(.*?)[+]{1}(.*)$".r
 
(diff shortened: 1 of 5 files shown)


app/utils/Utils.scala
```diff
@@ -23,6 +23,8 @@ import scala.util.Try
 object Utils {
   val acceptHeaderPattern = "^application/vnd[.]{1}hmrc[.]{1}(.*?)[+]{1}(.*)$".r
 
+  val acceptHeaderMissing = "The accept header is missing or invalid"
```
RobertLaverick, line 26: I'm not a fan of this living here, if we're going to be having stock messaging, we should probably isolate that. I'm also not sure if we have to support welsh here as its an API - I assume not, but at the least we can move this to a Messages object instead of Utils.

PR #74, opened 2020-05-22
CTDA-183 Added multiple mediaTypes to api definition

crystalmark:


resources/public/api/conf/1.0/application.raml
```diff
@@ -14,7 +14,7 @@ documentation:
  - title: Errors
    content: !include docs/errors.md
 
-mediaType: application/xml
```
RobertLaverick, line 17: quickly, why is this just media type and not the hmrc one?

crystalmark: Going by the RAML producers guide - this specifies the types that are accepted into the API but not the version.

PR #78, opened 2020-05-26
correctly mapping locations for ResponseArrivals

RobertLaverick:


app/models/response/ResponseArrival.scala
```diff
@@ -18,15 +18,16 @@ package models.response
...
 import play.api.libs.json.Json
+import utils.Utils
 
 object ResponseArrival {
 
   implicit val format = Json.format[ResponseArrival]
 
   def apply(a: Arrival): ResponseArrival = {
-    ResponseArrival(a.location, a.created, a.updated, a.movementReferenceNumber, a.status, a.messagesLocation)
+    ResponseArrival(s"/movements/arrivals/${Utils.urlEncode(a.arrivalId.toString)}", a.created, a.updated, a.movementReferenceNumber, a.status, s"/movements/arrivals/${Utils.urlEncode(a.arrivalId.toString)}/messages")
```
jhkldev, line 30: As it's a location, shouldn't the string start with "/customs/transits"? I think the get methods on the controllers  are missing this too.

app/models/response/ResponseArrival.scala
```diff
@@ -18,15 +18,16 @@ package models.response
...
-import models.domain.{Arrival}
+import models.domain.Arrival
 import play.api.libs.json.Json
+import utils.Utils
 
 object ResponseArrival {
 
   implicit val format = Json.format[ResponseArrival]
 
   def apply(a: Arrival): ResponseArrival = {
```
jhkldev, line 29: Can this method take in an extra argument for location and then you could pass in something like routes.ArrivalMovementController.getArrivals(arrivalId).url ? It's just to avoid hardcoding the path

app/models/response/ResponseArrival.scala
crystalmark: The /customs/transits prefix is only when we are in managed envs.  It's added by the API Platform so not needed for local testing.

jhkldev: > The /customs/transits prefix is only when we are in managed envs. It's added by the API Platform so not needed for local testing.

I think it's still required in the response here. That way external users can use the path directly in their code without having to add the /customs/transits prefix.

RobertLaverick: I won't lie, I'm against including it.
1) Because if we use the reverseRouting we don't get it.
2) Our root location is /customs/transits/ in the environments, and / locally. We don't include www.governmenturlstuff.com/ so why include the rest of the prelude?

PR #90, opened 2020-06-29
CTDA-344: Prevent lazy evaluation when creating map for supported message types in XSDFile

jhkldev:


app/models/request/XSDFile.scala
```diff
@@ -17,27 +17,34 @@
...
+    val Definitions = Seq(ArrivalNotificationXSD, UnloadingRemarksXSD)
+    val DefinitionsMap = Definitions.map(xsd => xsd.Label -> xsd).toMap
+    val SupportedMessages: Map[String, XSDFile] = DefinitionsMap.filter {
+      case (key: String, value: XSDFile) =>
+        value.isInstanceOf[ArrivalMessage]
+    }
+  }
 }
 
 object ArrivalNotificationXSD extends XSDFile {
```
crystalmark, line 37: Should ArrivalNotificationXSD also have the ArrivalMessage trait?

jhkldev: > Should ArrivalNotificationXSD also have the ArrivalMessage trait?

No as we don't want to allow the user to send an IE007 to the /messages endpoint.

PR #114, opened 2020-07-14
added GET /messages endpoint

RobertLaverick:


app/models/response/ResponseMessage.scala
```diff
@@ -18,17 +18,18 @@ package models.response
...
 import utils.NodeSeqFormat
 import utils.CallOps._
+
 import scala.xml.NodeSeq
 
 object ResponseMessage extends NodeSeqFormat{
   implicit val format = Json.format[ResponseMessage]
 
-  def apply(m: MovementMessage, arrivalId: String, messageId: String): ResponseMessage = ResponseMessage(routes.ArrivalMessagesController.getArrivalMessage(arrivalId, messageId).urlWithContext, m.dateTime, m.messageType, m.message)
+  def apply(m: MovementMessage, call: Call): ResponseMessage = ResponseMessage(call.urlWithContext, m.dateTime, m.messageType, m.message)
```
jhkldev, line 32: Can .urlWithContext be passed into MovementMessage?

I think the apply method should take in a single MovementMessage argument, with location being passed in to the MovementMessage constructor.

RobertLaverick: 1) I considered that, but this way we separate the presentation of the call from the call itself.
2) I can't do anything about using a single argument, since the movementmessage is the deserialized object that I'm mapping from that belongs to one of two backend apis

jhkldev: :ok_hand:

PR #119, opened 2020-07-20
Added GET /message/:messageId endpoint for departures

RobertLaverick:


it/connectors/DepartureMessageConnectorSpec.scala
```diff
@@ -0,0 +1,204 @@
...
+      val result = connector.getMessages("1").futureValue
+
+      result.isLeft mustEqual true
+      result.left.map { x => x.status mustEqual INTERNAL_SERVER_ERROR }
+    }
+  }
+
+  //TODO: Refactor this and other spec usages to a common trait
+  override protected def portConfigKey: String = "microservice.services.transits-movements-trader-at-departure.port"
+}
```
jhkldev, line 204: Can we move this to AppConfig for access, and specify the value in application.conf?

PR #124, opened 2020-08-06
Add public POST /departures/:departureId/messages endpoint

RobertLaverick:


app/controllers/actions/ValidateMessageAction.scala
```diff
@@ -34,8 +34,7 @@ class ValidateMessageAction @Inject()(xmlValidationService: XmlValidationService
       case body: NodeSeq =>
         if (body.nonEmpty) {
           val rootElementName = body.head.label
-
-          XSDFile.Arrival.SupportedMessages.get(rootElementName) match {
+          XSDFile.SupportedMessages.get(rootElementName) match {
```
jhkldev, line 37: This is incorrect as this will allow users to submit a request of release to the arrivals backend and unloading remarks to the departures backend.<br><br>There should be a ValidateArrivalMessageAction and ValidateDepartureMessageAction and the call should be XSDFile.Arrival.SupportedMessages.get(rootElementName) for arrivals and XSDFile.Departure.SupportedMessages.get(rootElementName) for departures.

jhkldev: There also needs to be tests for ValidateArrivalMessageAction and ValidateDepartureMessageAction.

app/models/request/XSDFile.scala
```diff
@@ -32,6 +33,17 @@ object XSDFile {
...
+  object Departure {
+    val Definitions = Seq(DepartureDeclarationXSD, RequestOfReleaseXSD)
+    val DefinitionsMap = Definitions.map(xsd => xsd.Label -> xsd).toMap
+    val SupportedMessages: Map[String, XSDFile] = DefinitionsMap.filter {
+      case (key: String, value: XSDFile) =>
+        value.isInstanceOf[DepartureMessage]
+    }
+  }
+
+  val SupportedMessages: Map[String, XSDFile] = Arrival.SupportedMessages ++ Departure.SupportedMessages
```
jhkldev, line 46: The XSDFile.SupportedMessages constant should be removed. See comment above.

PR #141, opened 2020-10-30
10,000 Euro Guarantee Action and Parser

RobertLaverick:


app/services/EnsureGuaranteeService.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  def ensureGuarantee(xml: NodeSeq): ParseHandler[NodeSeq] =
+    parseInstructionSets(xml) match {
+      case Left(error) => Left(error)
+      case Right(instructionSets) =>
+        Right(updateXml(prunedXml(xml), instructionSets))
+    }
+
+  def parseInstructionSets(xml: NodeSeq): ParseHandler[Seq[TransformInstructionSet]] = {
+    println("parseGuarantees"); xmlReaders.parseGuarantees(xml) match {
```
amaalali, line 42: Some cheeky debug `println`s have snuck in 🦗

app/utils/guaranteeParsing/DefaultGuaranteeApplier.scala
```diff
@@ -0,0 +1,35 @@
...
+    s.toDetails(g.gReference) match {
+      case Left(error) => Left(error)
+      case Right(details) => details.guaranteeAmount match {
+        case None => Right(details.copy(Some(BigDecimal(10000.00)), Some("EUR")))
+        case Some(amount) => amount match {
+          case a if a.equals(BigDecimal(0)) => Left(GuaranteeAmountZero("GuaranteeAmount cannot be zero"))
+          case _ => Right(details)
+        }
+      }
+    }
```
amaalali, line 34: You could simplify this with a `map` on line 25`s.toDetails(g.gReference) `

app/models/ParseError.scala
```diff
@@ -0,0 +1,47 @@
...
+  case class AdditionalInfoInvalidCharacters(message: String) extends ParseError
+  case class AdditionalInfoCodeMissing(message: String)       extends ParseError
+  case class CurrencyCodeInvalid(message: String)             extends ParseError
+  case class AmountStringInvalid(message: String)             extends ParseError
+  case class SpecialMentionNotFound(message: String)          extends ParseError
+  case class GuaranteeNotFound(message: String)               extends ParseError
+  case class MissingItemNumber(message: String)               extends ParseError
+  case class InvalidItemNumber(message: String)               extends ParseError
+
+  def liftParseError[A](input: Seq[Either[ParseError, A]]): ParseHandler[Seq[A]] =
```
amaalali, line 40: If we're using the alias for the return type, can we use it for the input type?

app/models/ParseHandler.scala
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package models
+
+trait ParseHandling {
+
+  type ParseHandler[A] = Either[ParseError, A]
+
+}
```
amaalali, line 23: This might sit better in the package object

app/models/SpecialMentionGuarantee.scala
```diff
@@ -0,0 +1,87 @@
...
+
+import models.ParseError.{AdditionalInfoInvalidCharacters, AdditionalInfoTooLong, AmountStringInvalid, CurrencyCodeInvalid}
+
+import scala.xml.NodeSeq
+
+trait SpecialMention
+
+case class SpecialMentionOther(xml: NodeSeq) extends SpecialMention
+
+case class SpecialMentionGuarantee(additionalInfo: String) extends SpecialMention {
```
amaalali, line 27: Should these be sealed traits with final case classes?

app/models/ParseError.scala
```diff
@@ -0,0 +1,47 @@
...
+    input.filterNot(i => i.isRight).headOption match {
+      case Some(error) => Left(error.left.get)
+      case None => Right(input.map {
+        x => x.right.get
+      })
+    }
```
amaalali, line 46: I believe this might be `sequence` in cats so we should be able to do `input.toList.sequence`. In light of that I think the name should be `sequenceErrors` since lift has the connotation (especially with Scala) of wrapping something in a context like Option or Either, but in this case were trying to "swap" the types Seq and ParseHandler, i.e. go from Seq[ParseHandler[A]] to ParseHandler[Seq[A]]

app/utils/guaranteeParsing/GuaranteeXmlReaders.scala
```diff
@@ -0,0 +1,104 @@
...
+    ParseError.liftParseError((xml \ "SPEMENMT2").map {
+      node =>
+        specialMention(node) match {
+          case Left(e) => Left(e)
+          case Right(s) => Right(s)
+        }
+    })
+  }
+
+  val gooBlock: ReaderT[ParseHandler, NodeSeq, Seq[GooBlock]] =
```
amaalali, line 51: I'm not sure if there are going to be multiple GooBlock, shouldn't there be just one?

amaalali: It looks like we're being forced to call head in `EnsureGuaranteeService` because we have a Seq[GooBlock] here

app/models/SpecialMentionGuarantee.scala
```diff
@@ -0,0 +1,87 @@
...
+
+
+case class SpecialMentionGuaranteeDetails(guaranteeAmount: Option[BigDecimal], currencyCode: Option[String], reference: String) {
+  def toSimple: SpecialMentionGuarantee = {
+    val amount = guaranteeAmount.getOrElse(BigDecimal(10000.00)).toString()
+    val currency = currencyCode.getOrElse("EUR")
+
+    SpecialMentionGuarantee(amount + currency + reference)
+  }
+}
```
amaalali, line 87: Should this be in it's own file?

app/services/EnsureGuaranteeService.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  def updateXml(prunedXml: NodeSeq, instructionSets: Seq[TransformInstructionSet]): NodeSeq = {
+    new RuleTransformer(new RewriteRule {
+      override def transform(node: Node): NodeSeq = {
+        node match {
+          case e: Elem if e.label == "GOOITEGDS" => {
+            xmlReaders.gooBlock(e) match {
+              case Left(_) => throw new Exception()
+              case Right(blocks) => {
+                val currentBlock = blocks.head
```
amaalali, line 66: I think this dangerous `head` might be because the parsers are wrong

app/utils/guaranteeParsing/GuaranteeXmlReaders.scala
RobertLaverick: We can have multiple GooBlocks in a single xml document. At least in the examples I've been given.

app/services/EnsureGuaranteeService.scala
RobertLaverick: Oh, I need to ensure this is a headOpt - I'll have to confirm the examples I have are legit, but I'll revisit this code.

app/utils/guaranteeParsing/DefaultGuaranteeApplier.scala
RobertLaverick: I tried this, but I was having issues with resolving the left further down. Could you show me how to resolve that problem?

app/models/ParseHandler.scala
RobertLaverick: Package object?

app/models/ParseError.scala
RobertLaverick: I'm confused, are you suggesting an implementation borrowing from cats or a simple rename?

app/models/SpecialMentionGuarantee.scala
```diff
@@ -0,0 +1,87 @@
...
+      Left(AmountStringInvalid(s"Invalid String for amount: $amountString $cutIndex $additionalInfo $guaranteeReference $currencyOpt"))
+    }
+  }
+}
+
+
+case class SpecialMentionGuaranteeDetails(guaranteeAmount: Option[BigDecimal], currencyCode: Option[String], reference: String) {
+  def toSimple: SpecialMentionGuarantee = {
+    val amount = guaranteeAmount.getOrElse(BigDecimal(10000.00)).toString()
+    val currency = currencyCode.getOrElse("EUR")
```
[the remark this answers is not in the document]
RobertLaverick: I must have overlooked this, I'd just seen I hadn't added tests for this when I relocated the class (I've pushed that change now) - so I'll add tests with coverage for this, as you're right about it only overriding when it doesn't exist.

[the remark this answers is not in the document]
RobertLaverick: @crystalmark just to check, I get that we want to override the currency if we set the amount - what do we want to do if we have an amount but no currency?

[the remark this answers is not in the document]
RobertLaverick: I've moved this conversion to the SpecialMentionGuarantee.toDetails method, since that handles the parsing of the value, but I've applied these comments and the error case there

app/models/SpecialMentionGuarantee.scala
```diff
@@ -0,0 +1,85 @@
...
+import scala.xml.NodeSeq
+
+sealed trait SpecialMention
+
+final case class SpecialMentionOther(xml: NodeSeq) extends SpecialMention
+
+final case class SpecialMentionGuarantee(additionalInfo: String) extends SpecialMention {
+  def toDetails(guaranteeReference: String) : Either[ParseError, SpecialMentionGuaranteeDetails] = {
+    if(additionalInfo.length > 18) return Left(AdditionalInfoTooLong("AdditionalInfo is too long"))
+    if(additionalInfo.matches("'[a-zA-Z0-9_.]*'")) return Left(AdditionalInfoInvalidCharacters("invalid characters in additional info"))
```
jhkldev, line 30: You could do a single if check here. A min and max length can be applied to the regex and the error message could be changed to "Additional information must contain only letters and numbers and must have a length of between 1 and 18 characters."<br><br>The error messages returned with Left in the PR could be made more human readable.

app/models/SpecialMentionGuarantee.scala
```diff
@@ -0,0 +1,85 @@
...
+              case (None, _) =>
+                Right(SpecialMentionGuaranteeDetails(BigDecimal(10000.00), "EUR", guaranteeReference))
+            }
+        }
+      }
+    }
+  }
+
+  private def getCurrencyCode(additionalInfo: String, guaranteeReference: String): Either[ParseError, Option[String]] = {
+    if(additionalInfo.length >= (guaranteeReference.length + 3))
```
jhkldev, line 52: Could additionalInfo and guaranteeReference be merged into a case class? Just to avoid/reduce the substring and length calls.

app/models/SpecialMentionGuarantee.scala
```diff
@@ -0,0 +1,85 @@
...
+      else if(additionalInfo.head.isDigit)
+      {
+        Right(None)
+      }
+      else Left(CurrencyCodeInvalid(s"Invalid Currency Code: $currencyCode"))
+    }
+    else Right(None)
+  }
+
+  private def getAmount(additionalInfo: String, guaranteeReference: String, currencyOpt: Option[String]): Either[ParseError, Option[BigDecimal]] = {
```
jhkldev, line 68: See comment on getCurrencyCode.

app/services/EnsureGuaranteeService.scala
```diff
@@ -0,0 +1,138 @@
...
+            }
+          }
+          case _ => node
+        }
+      }
+    }).transform(prunedXml)
+  }
+
+  def prunedXml(xml: NodeSeq): NodeSeq = {
+    new RuleTransformer(new RewriteRule {
```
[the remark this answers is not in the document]
RobertLaverick: Do you have an alternative?

app/models/SpecialMentionGuarantee.scala
RobertLaverick: You aren't wrong, but I wasn't sure merging it was worthwhile.

app/models/SpecialMentionGuarantee.scala
RobertLaverick: I'm not sure of the advantage here - is this just to isolate the string manipulation?

project/AppDependencies.scala
```diff
@@ -3,13 +3,17 @@ import sbt._
 
 object AppDependencies {
 
+  private val catsVersion = "2.1.1"
+
   val compile = Seq(
-    "uk.gov.hmrc"       %% "bootstrap-play-26"   % "1.13.0"
```
marcinkrykowski, line 7: Why was that lined removed & restored?

RobertLaverick: Because I added a comma to add the cats-core library

app/services/EnsureGuaranteeService.scala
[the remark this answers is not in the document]
RobertLaverick: I've tried to mimic a solution that is more performant, but had no luck, specifically with erasing nodes. It makes sense to create a technical debt ticket to address this though. @crystalmark are you happy with a tech debt ticket to improve xml transformation performance?

app/models/SpecialMentionGuarantee.scala
RobertLaverick: I'm resolving this as @jhkldev has said it was ok.

app/models/SpecialMentionGuarantee.scala
RobertLaverick: I'm resolving this as @jhkldev has said it was ok

app/models/SpecialMentionGuarantee.scala
RobertLaverick: I'm resolving this as @jhkldev has said it was ok

app/services/EnsureGuaranteeService.scala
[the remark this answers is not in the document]
crystalmark: Happy with a tech dept ticket but we should compare the before and after performance runs to make sure this does not have a significant hit.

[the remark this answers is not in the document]
RobertLaverick: Technical Debt Ticket : https://jira.tools.tax.service.gov.uk/browse/CTDA-898

Resolving

jhkldev (approved):

marcinkrykowski (approved):

crystalmark (approved):

crystalmark (approved):

PR #147, opened 2020-11-09
ensured special mentions themselves get changed rather than removed a…

RobertLaverick:
…nd reinserted afterwards

### DIFF
--- a/app/services/EnsureGuaranteeService.scala
+++ b/app/services/EnsureGuaranteeService.scala
@@ -35,7 +35,7 @@ class EnsureGuaranteeService @Inject()(xmlReaders: GuaranteeXmlReaders, instruct
     parseInstructionSets(xml) match {
       case Left(error) => Left(error)
       case Right(instructionSets) =>
-        Right(updateXml(prunedXml(xml), instructionSets))
+        Right(updateXml(xml, instructionSets))
     }
 
   def parseInstructionSets(xml: NodeSeq): ParseHandler[Seq[TransformInstructionSet]] = {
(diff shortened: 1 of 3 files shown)


app/services/EnsureGuaranteeService.scala
```diff
@@ -55,58 +55,43 @@ class EnsureGuaranteeService @Inject()(xmlReaders: GuaranteeXmlReaders, instruct
...
 
-  def updateXml(prunedXml: NodeSeq, instructionSets: Seq[TransformInstructionSet]): NodeSeq = {
+  def updateXml(originalXml: NodeSeq, instructionSets: Seq[TransformInstructionSet]): NodeSeq = {
     new RuleTransformer(new RewriteRule {
       override def transform(node: Node): NodeSeq = {
         node match {
           case e: Elem if e.label == "GOOITEGDS" => {
             xmlReaders.gOOITEGDSNodeFromNode(e) match {
-              case Left(_) => throw new Exception()
+              case Left(_) => throw new Exception("Unable to parse GOOITEGDSNode on update")
```
jhkldev, line 64: Can we return Option[NodeSeq] here and then pattern match further up the call stack? We want to avoid throwing exceptions.

RobertLaverick: Sadly I don't think we can - we're already basically at the top of the stack and transform has a NodeSeq signature. Otherwise I wouldn't be throwing an exception here.

It shouldn't ever throw however as this is the second time this xml should be going through this sort of parser and should be safe. I appreciate the reasoning but without a complete rewrite of RewriteRule, or an alternative to override, I'm stuck with this exception, and the other on line 71

jhkldev: > Sadly I don't think we can - we're already basically at the top of the stack and transform has a NodeSeq signature. Otherwise I wouldn't be throwing an exception here.
> 
> It shouldn't ever throw however as this is the second time this xml should be going through this sort of parser and should be safe. I appreciate the reasoning but without a complete rewrite of RewriteRule, or an alternative to override, I'm stuck with this exception, and the other on line 71

We could return null instead of throwing an exception and then check against null.

RobertLaverick: At which point we obfuscate the error that has happened in exchange for a processing saving. That's not great.

jhkldev (approved):

crystalmark (approved):

PR #148, opened 2020-11-12
Xml to Json

GeorgeColdham:
#XML 2 JSON

Helper to transform XML to JSON format

transformation

### DIFF
--- a/test/utils/JsonHelperSpec.scala
+++ b/test/utils/JsonHelperSpec.scala
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2020 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


test/utils/JsonHelperSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+
+package utils
+
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+import play.api.libs.json.{JsObject, Json}
+
+class JsonHelperSpec extends AnyFreeSpec with Matchers {
+
+  "JsonHelper" - {
```
RobertLaverick, line 25: can you add an example of a full message?

RobertLaverick (changes requested):

GeorgeColdham:
New tests added, will look into why the parser seems to change the order of the JSON and if it even matters.

crystalmark (approved):

PR #154, opened 2020-11-26
Added parsers for OthRefREF4 support to allow guarantees other than 1,2,4 or 9

RobertLaverick:


crystalmark (approved):

marcinkrykowski (approved):

app/models/Guarantee.scala
```diff
@@ -17,3 +17,9 @@
 package models
 
 case class Guarantee(gType: Int, gReference: String)
+
+object Guarantee {
+  val referenceTypes = Seq[Int](0, 1, 2,4, 9)
```
jhkldev, line 22: Can we make the reference types be an enum/sealed trait?

RobertLaverick: I don't know what the types are to name them. I'm not sure we have any need for any of the individual ones either.

TheNickWilson: You can find the names in the ref data file if you do want to use them (up to you if you think they'd aid clarity of course!) 
 They are:

0 - Guarantee waiver
1 - Comprehensive guarantee
2 - Individual guarantee
3 - Cash deposit guarantee
4 - Flat-rate voucher
5 - Guarantee waiver, secured amount not greater than 500 EUR
6 - Guarantee waived
7 - Guarantee not required for the journey between OoDep and OoTra
8 - Guarantee not required for certain public bodies
9 - Individual Guarantee with multiple usages
A - Guarantee waiver by agreement
B - TIR

RobertLaverick: I've converted the uses of type to a char to account for A and B, rather than a full enum/sealed-trait solution since we have no use for the individual types.

jhkldev (approved):

PR #159, opened 2020-12-09
CTDA-977 streamlined value defaulting and ensure GB-to-GB doesn't default

RobertLaverick:


app/models/Offices.scala
```diff
@@ -16,15 +16,7 @@
 
 package models
 
-import org.scalatest.freespec.AnyFreeSpec
-import org.scalatest.matchers.must.Matchers
-import org.scalatestplus.mockito.MockitoSugar
+abstract class Office(value: String)
```
crystalmark, line 19: Weird how git thinks that SpecialMentionGuaranteeDetailsSpec.scala 
 has been moved to app/models/Offices.scala

app/models/SpecialMentionGuarantee.scala
```diff
@@ -74,14 +67,14 @@ final case class SpecialMentionGuarantee(additionalInfo: String) extends Special
...
-      Right(Some(BigDecimal(amountString)))
+    else if(amountString.length >= 16) {
+      Left(AmountStringTooLong("Amount is longer than 16 characters"))
     }
     else {
-      Left(AmountStringInvalid(s"Invalid String for amount: $amountString $cutIndex $additionalInfo $guaranteeReference $currencyOpt"))
-    }
+      Try(BigDecimal(amountString)) match {
+        case Success(bigDecimal) => Right(Some(bigDecimal))
+        case Failure(_) =>  Left(CurrencyCodeInvalid("Currency Code Invalid"))
```
crystalmark, line 76: Should this be CurrencyCodeInvalid or something like AmountIsNotAValidNumber ??

app/utils/guaranteeParsing/RouteChecker.scala
```diff
@@ -0,0 +1,43 @@
...
+  def gbOnlyCheck(xml: NodeSeq): Either[ParseError, Boolean] =
+    xmlReaders.officeOfDeparture(xml) match {
+      case Left(error) => Left(error)
+      case Right(departure) => {
+        xmlReaders.officeOfDestination(xml) match {
+          case Left(error) => Left(error)
+          case Right(destination) => {
+            (departure.value.startsWith("GB"), destination.value.startsWith("GB")) match {
+              case (true, destinationGB) => Right(destinationGB)
+              case (false, _) => Left(InappropriateDepartureOffice("Inappropriate Departure Office"))
```
jhkldev, line 36: An if statement would look clearer here.

`if (departure.value.startsWith("GB") && destination.value.startsWith("GB")) Right(true) else Left(InappropriateDepartureOffice("Inappropriate Departure Office"))`

Does the check need to be case sensitive?

Could the method signature be changed to return a Boolean? With ParseError created higher up the call stack. The method name could also be changed to `isGBOnly`.

it/services/EnsureGuaranteeServiceIntegrationSpec.scala
```diff
@@ -291,6 +290,8 @@ class EnsureGuaranteeServiceIntegrationSpec extends AnyFreeSpec with Matchers wi
 
       val result = service.ensureGuarantee(standardInputXML)
 
+      println(result)
```
jhkldev, line 293: Needs to be removed.

it/services/EnsureGuaranteeServiceIntegrationSpec.scala
```diff
@@ -301,6 +302,8 @@ class EnsureGuaranteeServiceIntegrationSpec extends AnyFreeSpec with Matchers wi
...
```
jhkldev, line 305: Needs to be removed.

app/services/EnsureGuaranteeService.scala
```diff
@@ -16,26 +16,26 @@
...
 
 
   def ensureGuarantee(xml: NodeSeq): ParseHandler[NodeSeq] =
-    parseInstructionSets(xml) match {
+    routeChecker.gbOnlyCheck(xml) match {
       case Left(error) => Left(error)
-      case Right(instructionSets) =>
-        Right(updateXml(xml, instructionSets))
+      case Right(gbOnly) if gbOnly => Right(xml)
+      case _ =>
```
jhkldev, line 33: `case Right(_) =>`

jhkldev (changes requested):

app/models/Offices.scala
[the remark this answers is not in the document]
RobertLaverick: Yes, but I wanted to enforce the value field without having to do something odd like having value as a def. Not sure a trait is better.

app/utils/guaranteeParsing/RouteChecker.scala
RobertLaverick: I need to use the value of destination.value.startsWith so that if statement wouldn't work, but another one would.

No, I need the ParseError Either because the two mapped xmlReader methods both return ParseErrors and those values need to be handled

it/services/EnsureGuaranteeServiceIntegrationSpec.scala
RobertLaverick: Good spot

it/services/EnsureGuaranteeServiceIntegrationSpec.scala
RobertLaverick: Good Spot

app/models/SpecialMentionGuarantee.scala
```diff
@@ -74,14 +67,14 @@ final case class SpecialMentionGuarantee(additionalInfo: String) extends Special
     {
       Right(None)
     }
-    else if(amountString.length > 18) {
-      Left(AmountStringTooLong("Amount is longer than 18 characters"))
-    }
-    else if(amountString.matches("^[0-9]*\\.[0-9]{2}$")) {
-      Right(Some(BigDecimal(amountString)))
+    else if(amountString.length >= 16) {
+      Left(AmountStringTooLong("Amount is longer than 16 characters"))
```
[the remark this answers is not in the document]
RobertLaverick: I've changed it to "Guarantee amount should not be greater than 15 characters", because that reflects the check

jhkldev (approved):

app/models/SpecialMentionGuarantee.scala
```diff
@@ -75,13 +68,13 @@ final case class SpecialMentionGuarantee(additionalInfo: String) extends Special
...
-    else if(amountString.matches("^[0-9]*\\.[0-9]{2}$")) {
-      Right(Some(BigDecimal(amountString)))
+      Left(AmountStringTooLong("Guarantee amount should be not be longer than 18 characters"))
     }
     else {
-      Left(AmountStringInvalid(s"Invalid String for amount: $amountString $cutIndex $additionalInfo $guaranteeReference $currencyOpt"))
-    }
+      Try(BigDecimal(amountString)) match {
+        case Success(bigDecimal) => Right(Some(bigDecimal))
+        case Failure(_) =>  Left(AmountStringInvalid("Amount String Invalid"))
```
jhkldev, line 76: The error message could be changed to "The specified amount is not a valid representation of a decimal." (to reflect the NumberFormatException being thrown)

jhkldev (changes requested):

jhkldev (approved):

crystalmark (approved):

PR #162, opened 2020-12-17
CTDA-817: Only send GuaranteeAdded audit when guarantee is added

TheNickWilson:


RobertLaverick (approved):
:shipit: (thank you, this was on my mind all week) (Also don't mind me, the Opencast Party is in 3 mins so I thought I'd do a little review)

PR #175, opened 2021-02-11
spike to add ni-function

RobertLaverick:
updates the 2020-2021 docs
also updates the routechecker to return a right of false when an XI Departure office is provided so that the guarantee will be applied

### DIFF
--- a/app/utils/guaranteeParsing/RouteChecker.scala
+++ b/app/utils/guaranteeParsing/RouteChecker.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 HM Revenue & Customs
+ * Copyright 2021 HM Revenue & Customs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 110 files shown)


app/utils/guaranteeParsing/RouteChecker.scala
```diff
@@ -31,15 +31,13 @@ class RouteChecker @Inject()(xmlReaders: GuaranteeXmlReaders) {
...
           case Left(error) => Left(error)
           case Right(destination) => {
-            if(departure.value.startsWith("GB")){
-              Right(destination.value.startsWith("GB"))
-            }
-            else {
-              Left(InappropriateDepartureOffice("Inappropriate Departure Office"))
+            departure.prefix match {
+              case "GB" => Right(destination.value.startsWith("GB"))
+              case "XI" => Right(false)
```
crystalmark, line 36: Did we decide that there are no invalid ports, anything that's not GB or XI will go to GB?  Did I dream it?

RobertLaverick: I've no point of reference. This only focuses on the departure office, and we don't handle anything that isn't XI or GB if that's what you're questioning.

If you're thinking about arrivals, that isn't relevant to the routechecker.

crystalmark (approved):

PR #184, opened 2021-04-01
Guarantee spike

RobertLaverick:


GeorgeColdham (approved):
LGTM although given its size etc, I would wait for @crystalmark to take a look before merging.

crystalmark (approved):

PR #187, opened 2021-04-16
CTDA-1281 Add metrics collection to controllers and connectors

DavidGregory084:
Add metrics collection to the controllers and connectors in the API

I have used the convention of using the controller method name in our controllers, and in using `<movement type>-backend-<http method>` in the connectors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,8 @@ test-result
 tmp
 .idea_modules
 bin
+.bloop/
+.bsp/
+.metals/
+.vscode/
+metals.sbt
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,44 @@
+version = "2.7.5"
+
+// Sets max amount of characters before new line
+maxColumn = 160
+
+// Sets indentation of method arguments and parameters
+continuationIndent {
+  callSite = 2
+  defnSite = 2
+}
+
+// Aligns tokens for numerous methods e.g. matchers, for yield and module ID's
+align = more
+
+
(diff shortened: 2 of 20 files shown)


RobertLaverick (changes requested):
Looks good, one small change - can you change the metric params from magic strings to something defined somewhere?

DavidGregory084:
I think with magic numbers and magic strings, usually we don't like them because they are not self-evident, so it's not apparent why that value was chosen, and they prevent us from changing a value throughout our program if they are reused in multiple places.

I am open to the change but it's additional indirection when searching for the occurrence of a metric - these strings only occur in one place, must be unique and are (hopefully) self-explanatory. Do you think we will get a lot of value out of extracting them as variables @RobertLaverick?

RobertLaverick:
I appreciate the sentiment and its totally sensible. However I'm just cautious of accidental changes to the strings, or the normalisation of magic strings in repos. We're all human and its easy to miss a single character change by accident in a PR, but that then means broken metrics. Further it means we can keep all our metric names in one place, which acts as self-documentation of metrics that we use, as well as silo-ing off the strings so they can't be accidentally changed (and that is exactly the sort of thing that can fail more quietly than we'd expect)

DavidGregory084:
Ok @RobertLaverick I've addressed that - lmk what you think

RobertLaverick (approved):

DavidGregory084:
@RobertLaverick could you have another quick look? I have added histograms to document the number of arrivals, departures and corresponding messages that are returned by each endpoint. I also have refactored HasMetrics a little bit as some of the code was copied from another project (LocalMetrics) etc. and was not really needed.

RobertLaverick (approved):

PR #194, opened 2021-05-06
CTDA-1271 added security token and notification url field definitions

RobertLaverick:


crystalmark (approved):

resources/public/api/definition.json
```diff
@@ -18,7 +18,18 @@
         "endpointsEnabled": true,
         "access": {
           "type": "PUBLIC"
-        }
+        },
+        "fieldDefinitions": [{
+          "name": "notificationUrl",
+          "description": "What is your notification web address for us to send push notifications to?",
+          "type": "URL",
```
DavidGregory084, line 25: This should be a PPNSField

resources/public/api/definition.json
```diff
@@ -18,7 +18,18 @@
...
+          "hint": "You must only give us a web address that you own. Your application will use this address to listen to notifications from HMRC."
+        }, {
```
DavidGregory084, line 27: I think there should be validation rules for this field similar to https://github.com/hmrc/goods-movement-system-haulier-api/blob/master/app/uk/gov/hmrc/goodsmovementsystemhaulierapi/platform/views/definition.scala.txt#L30

DavidGregory084 (changes requested):

resources/public/api/definition.json
DavidGregory084: Again see e.g. https://github.com/hmrc/goods-movement-system-haulier-api/blob/master/app/uk/gov/hmrc/goodsmovementsystemhaulierapi/platform/views/definition.scala.txt#L30

DavidGregory084 (approved):
Looks good to me 💯

PR #196, opened 2021-05-07
Ctda 1309

RobertLaverick:


resources/public/api/conf/1.0/application.raml
```diff
@@ -383,6 +385,12 @@ traits:
       is:
         - headers.acceptHeader
         - acceptHeaderInvalid
+       queryParameters:
+         updatedSince:
+           required: false
+           type: datetime
+           description: This is an optional filtering parameter. It allows you to request only the Movement Arrivals which have been updated since the date and time you have given us.
```
DavidGregory084, line 392:
```suggestion
           description: This is an optional filtering parameter. It allows you to request only the Movement Departures which have been updated since the date and time you have given us.
```

DavidGregory084 (approved):
Looks good, just one amendment that I think we should make in the API doc 👍

PR #208, opened 2021-06-18
WIP: Refactoring guarantee parsing

DavidGregory084:
* Using some more Either functions and helper functions rather than pattern matching
* The magic of traverse

Suggest reviewing with whitespace changes hidden!


drnaylor-oc:
Lots has changed since this was opened and as this part of the code is approaching the end of its life (and it's working), this isn't going to happen now -- as nice as it would have been.

Closing this PR.

PR #215, opened 2021-07-23
Ctda 1475/json errors

GeorgeColdham:
# Json Errors

[Ticket](https://jira.tools.tax.service.gov.uk/browse/CTDA-1475)


RobertLaverick (approved):
Nothing game-breaking, but it would be nice if the pattern of 
`val errorResponse = XmlParseJsonErrorResponse.fromParseError(error)
            Future.successful(Left(BadRequest(Json.toJson(errorResponse))))`
            and
            `          val errorResponse = XmlParseJsonErrorResponse(XmlError.RequestBodyEmptyMessage)
          Future.successful(Left(BadRequest(Json.toJson(errorResponse))))`
            
            could be simplified into one line - perhaps with an alias import so you can just do xpjer(error) instead?

PR #221, opened 2021-09-15
Ctda 1600

RobertLaverick:
Added GetBoxInfo endpoint

### DIFF
--- a/app/controllers/PushPullNotificationController.scala
+++ b/app/controllers/PushPullNotificationController.scala
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2021 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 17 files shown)


app/controllers/PushPullNotificationController.scala
```diff
@@ -0,0 +1,54 @@
...
+      request.headers.get(Constants.XClientIdHeader) match {
+        case Some(clientId) => pushPullNotificationConnector.getBox(clientId).map {
+          response =>
+          response match {
+            case Left(error) if(error.statusCode == NOT_FOUND) => NotFound("No box found for your client id")
+            case Left(_) => InternalServerError("Unexpected Error")
+            case Right(box) => Ok(Json.toJson(box))
+          }
+        }
+        case None =>  Future.successful(BadRequest("Client Id Required"))
```
DavidGregory084, line 50: We should use JSON errors in all the responses here

RobertLaverick: So essentially just wrap it in a JsString like `BadRequest(JsString("Client..."))` ?

DavidGregory084: No - the errors from our API need to follow the format from the Developer Hub [Reference Guide](https://developer.service.hmrc.gov.uk/api-documentation/docs/reference-guide#errors) - so they need to have a `code` and a `message` field. See e.g. https://github.com/hmrc/common-transit-convention-traders/blob/master/app/models/response/JsonErrorResponse.scala

DavidGregory084: Or https://github.com/hmrc/common-transit-convention-guarantee-balance/blob/master/app/models/errors/BalanceRequestError.scala

DavidGregory084 (approved):

PR #223, opened 2021-09-23
Ctda 1600 fix raml

RobertLaverick:
I've two commits, one with and one without a Hateoas response because it slipped through the next last time. I can revert the Hateoas commit if we want to keep it simple

### DIFF
--- a/app/models/response/HateoasResponseBox.scala
+++ b/app/models/response/HateoasResponseBox.scala
@@ -27,7 +27,6 @@ object HateoasResponseBox {
 
     val boxUrl = routes.PushPullNotificationController.getBoxInfo().urlWithContext
 
-    JsObject(
       Json
         .obj(
           "boxId"                      -> box.boxId,
@@ -36,10 +35,5 @@ object HateoasResponseBox {
             "self"     -> Json.obj("href" -> boxUrl),
           )
         )
-        .fields
-        .filter(
-          t => t._2 != JsNull
-        )
-    )
   }
 }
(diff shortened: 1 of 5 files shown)


resources/public/api/conf/1.0/application.raml
```diff
@@ -643,3 +643,23 @@ traits:
                       }
               404:
                 description: When message does not exist, has been archived or is not available to the EORI number.
+  /push-pull-notifications/box:
+    get:
+      displayName: Get Box Info
+      description: |
+        Get Box Information for your application if there is a push-pull-notification box
```
DavidGregory084, line 650: Does this need content review?

app/models/response/HateoasResponseBox.scala
```diff
@@ -0,0 +1,45 @@
...
+        .obj(
+          "boxId"                      -> box.boxId,
+          "boxName"                 -> box.boxName,
+          "_links" -> Json.obj(
+            "self"     -> Json.obj("href" -> boxUrl),
+          )
+        )
+        .fields
+        .filter(
+          t => t._2 != JsNull
```
DavidGregory084, line 41: Can any of these things be null?

DavidGregory084 (approved):
Looks good to me, just a couple questions that don't necessarily need to block merging to QA

app/models/response/HateoasResponseBox.scala
RobertLaverick: No, its just a leftover, I'll clean it out.

PR #224, opened 2021-09-23
CTDA-1645 ensure that Play SecurityHeadersFilter is enabled

DavidGregory084:


conf/application.conf
```diff
@@ -35,22 +35,19 @@ play.modules.enabled += "uk.gov.hmrc.play.bootstrap.AuditModule"
 # A metric filter must be provided
 play.modules.enabled += "uk.gov.hmrc.play.bootstrap.graphite.GraphiteMetricsModule"
 
-# Provides an implementation and configures all filters required by a Platform frontend microservice.
-play.modules.enabled += "uk.gov.hmrc.play.bootstrap.MicroserviceModule"
-play.http.filters = "uk.gov.hmrc.play.bootstrap.filters.MicroserviceFilters"
```
DavidGregory084, line 40: This line is currently disabling all other filters by replacing the configuration value

The correct way to do this as of Play 2.6+ (I think) is to do `play.filters.enabled += WhateverFilter`, but the correct filters should already be enabled by the include of `backend.conf` at the top of this file.

conf/application.conf
```diff
@@ -35,22 +35,19 @@ play.modules.enabled += "uk.gov.hmrc.play.bootstrap.AuditModule"
...
+# Provides an implementation and configures all filters required by a Platform backend microservice.
+play.modules.enabled += "uk.gov.hmrc.play.bootstrap.backend.BackendModule"
```
DavidGregory084, line 39: This is a more specific set of configuration for a backend service

RobertLaverick (approved):

PR #235, opened 2022-02-15
CTDA-1712

RobertLaverick:
spike to prove we can produce multiple error messages - has limitations

Current state cannot list multiple errors involving same-level siblings.

i.e. we can't list three seperate parsing errors for the top level xml being missing, but we can list one, and then the first missing xml element in all child elements (i.e. 1 from HEAHEA, 1 from TRACONCE1, 1 from CUSOFFTRARNS)

I think that is a limitation of the parser in use as we are effectively capturing the first


drnaylor-oc (approved):
I don't see anything untoward about this - looks good to me.

PR #236, opened 2022-03-09
CTDA 1861: Defensively remove namespaces from XML root

drnaylor-oc:
Not 100% sure about this yet, need to check a couple of things and it could be quite slow with a large XML tree that has been namespaced -- not sure we want that. If we just want to check for the standard XML namespace, I may try a different approach involving strings and regex - which will require a different tack regardless.

Scalafmt also had its way, so some diffs are somewhat irrelavant.


test/utils/XmlHelperSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+          <child>test</child>
+        </root>,
+        <root xmlns:xsd="test">
+          <child>test</child>
+        </root>,
+        <root xmlns:xsd="test" xmlns:xsi="test">
+          <child>test</child>
+        </root>
+      )
+
```
liandra-giacomine, line 51: Is it worth adding an attribute to the child element to ensure the recursion is working as expected?

liandra-giacomine (approved):

test/utils/XmlHelperSpec.scala
drnaylor-oc: Discussed this in mob, but to summarise - yes and no. It's a good idea if we _also_ have a root element with a namespace on it, as that'll be an explicit verification of the recursive algorithm, so I'll add that! However, not in the case of if only the child element has it, in the large message trees we get, I don't think we want to traverse every single message we get - I don't think it's likely we'll get a message that has a namespace on a child only.

RobertLaverick (approved):

drnaylor-oc:
Closing in favour of #238

drnaylor-oc:
Closing in favour of #238

PR #243, opened 2022-04-06
CTCP-182: Submit a Departure Declaration XML

liandra-giacomine:
Route departure POST request according to accept header

### DIFF
--- a/app/controllers/DeparturesController.scala
+++ b/app/controllers/DeparturesController.scala
@@ -35,7 +35,7 @@ import models.response.HateoasDeparturePostResponseMessage
 import models.response.HateoasResponseDeparture
 import models.response.HateoasResponseDepartures
 import play.api.libs.json.Json
-import play.api.mvc.{Action, AnyContent, ControllerComponents, Request}
+import play.api.mvc.{Action, AnyContent, ControllerComponents, Request, Result}
 import uk.gov.hmrc.http.HttpErrorFunctions
 import uk.gov.hmrc.play.bootstrap.backend.controller.BackendController
 import utils.CallOps._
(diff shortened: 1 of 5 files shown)


app/controllers/DeparturesController.scala
```diff
@@ -62,13 +60,17 @@ class DeparturesController @Inject() (
...
+    with VersionSwitch {
 
   import MetricsKeys.Endpoints._
 
   lazy val departuresCount = histo(GetDeparturesForEoriCount)
 
   def submitDeclaration(): Action[NodeSeq] =
+    versionSwitch(submitDeclarationVersionOne, submitDeclarationVersionTwo)
+
+  def submitDeclarationVersionOne(): Action[NodeSeq] =
```
derekspalmer, line 73: private method?

derekspalmer: same for versionTwo.

app/controllers/VersionSwitch.scala
```diff
@@ -0,0 +1,41 @@
...
+    Action.async(parse.xml) {
+    (request: Request[NodeSeq]) =>
+      request.headers.get("accept") match {
+        case Some("application/vnd.hmrc.2.0+json") => v2Action(request)
+
+        case Some("application/vnd.hmrc.1.0+json") => v1Action(request)
+
+        case None => v1Action(request)
+
+        case Some(headerVal) => Future.successful(UnsupportedMediaType(s"Unsupported Accept-header: $headerVal"))
```
derekspalmer, line 38: Is it worth adding logging here?

derekspalmer (approved):
LGTM

app/controllers/VersionSwitch.scala
```diff
@@ -0,0 +1,41 @@
...
+import scala.concurrent.Future
+import scala.xml.NodeSeq
+import uk.gov.hmrc.play.bootstrap.backend.controller.BackendController
+
+trait VersionSwitch {
+ self: BackendController =>
+
+  def versionSwitch(v1Action: Action[NodeSeq], v2Action: Action[NodeSeq]): Action[NodeSeq] =
+    Action.async(parse.xml) {
+    (request: Request[NodeSeq]) =>
```
derekspalmer, line 30: type superfluous.

drnaylor-oc: It's not necessarily. When originally building this, we were having type inference issues causing the compiler to pick the wrong overload to `Action#apply`, as a `Request[_]` is also a `RequestHeader`. I haven't checked it here, but I'd imagine that it's possible the same thing is happening here.

This is a case where I think I'd prefer to keep it to ensure the compiler does the right thing -- even if removing it now works for some reason.

derekspalmer: Not sure about that - if the inferencer can't determine the type, it will let us know!
But ok - leave it.

drnaylor-oc: In this case, it wasn't that it couldn't infer a type, it's that it infered a super type of what it actually was - we needed to correct that.

drnaylor-oc (approved):

drnaylor-oc (approved):

PR #255, opened 2022-05-11
CTCP-266: Add validation connector

drnaylor-oc:
Enables a message to be validated via the Validation Service.


RobertLaverick (commented):
Everything looks good so far but there are a lot of changes mixed in here that might have been better as small PRs on their own. It looks like you've committed well to be able to cherry pick those changes, but if you want to get some of those smaller changes merged first, make those PRs and we can review them with that focus

drnaylor-oc:
Yeah, I got carried away it just sorting it all out. I'll see what I can do.

drnaylor-oc:
Actually: https://github.com/hmrc/common-transit-convention-traders/pull/251 could be reviewed as that forms the basis of this (it's the CTCP-304 commits). 

I'll need to split out the BaseError and the v2 package movement in the last commit, but I don't see that as too hard to do.

drnaylor-oc:
I've updated this PR to be the main bulk of the ticket changes. This is in draft for two reasons:

* The first commit is also PR #257. It's an important change that I just want to have a final discussion on.
* This will change slightly once CTCP-258 is done.

Also of note, because it's not available in bootstrap-play yet and there are classpath issues by including http-verbs 14.1, we can't use streaming via that, instead having to use the WSClient which means we can't get implicit auditing right now.

Beyond that - this should get some initial review.

RobertLaverick (changes requested):

drnaylor-oc:
I have this working with hmrc/transit-movements-validator#4 - it now handles the validation correctly. An example of an error message returned from the validator:

```json
{
    "message": "Request failed schema validation",
    "code": "SCHEMA_VALIDATION",
    "validationErrors": [
        {
            "lineNumber": 1,
            "columnNumber": 1,
            "message": "Content is not allowed in prolog."
        }
    ]
}
```

RobertLaverick (approved):

PR #257, opened 2022-05-11
CTCP-266: Move v2 code into v2 package and add departures router to separate routing from the v1 code

drnaylor-oc:
**Depends on #251 and #256**

Once the above two PRs are merged, only the last commit will be relavant for this PR. This moves all v2 code into its own package and separates out the routing code into its own controller (`DepraturesRouter`).

This will enable writing the Validation connector in the correct place.

Will rebase once the two PRs above are merged (or fates otherwise decided)

### DIFF
--- a/app/v2/controllers/V2DeparturesController.scala
+++ b/app/v2/controllers/V2DeparturesController.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2022 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package v2.controllers
+
+import akka.stream.Materializer
+import akka.stream.scaladsl.Sink
+import akka.stream.scaladsl.Source
+import akka.util.ByteString
+import com.google.inject.ImplementedBy
+import com.google.inject.Inject
+import com.google.inject.Singleton
+import play.api.Logging
+import play.api.mvc.Action
+import play.api.mvc.BaseController
+import play.api.mvc.ControllerComponents
+import routing.VersionedRouting
+import v2.controllers.actions.AuthNewEnrolmentOnlyAction
+import v2.controllers.actions.MessageSizeAction
+import v2.controllers.stream.StreamingParsers
+
+import scala.concurrent.Future
+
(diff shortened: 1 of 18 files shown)


liandra-giacomine (approved):
LGTM

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -0,0 +1,69 @@
...
+import play.api.mvc.ControllerComponents
+import routing.VersionedRouting
+import v2.controllers.actions.AuthNewEnrolmentOnlyAction
+import v2.controllers.actions.MessageSizeAction
+import v2.controllers.stream.StreamingParsers
+
+import scala.concurrent.Future
+
+@ImplementedBy(classOf[V2DeparturesControllerImpl])
+trait V2DeparturesController {
```
derekspalmer, line 38: Rather than 2 distinct V1, V2 Traits could we have 1 generic as they have the same signature, eg:
 a DepartureController[T] {
def submitDeclaration(): Action[T]
}

derekspalmer (approved):

app/v2/controllers/V2DeparturesController.scala
drnaylor-oc: No - just because `submitDeclration()` might be the same in both, we might end up with routes that will only exist in one or the other controller, or potentially have different parameters - so I'd rather not do that.

drnaylor-oc:
Anyone else? This is a structural change so I just want to make sure people are happy.

(Also, can't merge this right now, smoke tests are failing due to the test support API, so while this would build, tests would fail and so this won't deploy)

RobertLaverick (changes requested):
In principle what I've seen I like.

What I would ask is that if we're going to have a 'v2' package at the app root level, lets also have a 'v1' package and ensure that we've a clean package structure from the offset. Anything not explicitly v1 or v2 could be put into a 'common' package but I'm less fussed by that as it's a reasonable default. (For example, the CallOps object will be useful for both, and the routing classes aren't version specific either).

Connectors I'm less concerned about being V1'd or V2'd (though I expect there won't be much overlap, there is going to be some, like Push Pull) since the majority of the backend is new for v2.

drnaylor-oc:
So, I had a look at this and it's not nearly as simple as moving directories in the IDE unfortunately. Many errors appear by just renaming, and it turns out that Play has a lot of conventions that the IDE doesn't account for.

I'm not against that sort of change in principle, but the following issues exist:

* The API has, until now, followed all of Play's conventions to make things like writing the routes file easier, for example, query objects are in `models.domain`. Moving stuff out of there means we have to sort out the routes file (though we do anyway because of moving things about)
* We may lose the file history because of the renames
* My IDE does as half hearted job and there are a lot of errors that I need to pick up.

Given that this job is a lot more involved than just moving files, I'm tempted to say this should be a tech debt ticket for picking up by someone else (or me, but later). The reason I did this for v2 now is because I needed the routing to be done anyway, and there wasn't much to move. For v1 stuff, this is a larger task than expected - but isn't a dependency for this task.

Does that sound reasonable to you @RobertLaverick?

RobertLaverick (approved):

PR #262, opened 2022-05-26
CTCP-326: Connect API layer to transit-movements

drnaylor-oc:
Very WIP, feedback welcome as this progresses.

### DIFF
--- a/app/v2/controllers/V2DeparturesController.scala
+++ b/app/v2/controllers/V2DeparturesController.scala
@@ -39,13 +39,16 @@ import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.http.HeaderCarrierConverter
 import v2.controllers.actions.AuthNewEnrolmentOnlyAction
 import v2.controllers.actions.MessageSizeAction
+import v2.controllers.request.AuthenticatedRequest
 import v2.controllers.stream.StreamingParsers
 import v2.models.errors.BaseError
 import v2.models.errors.FailedToValidateError
 import v2.models.errors.FailedToValidateError.InvalidMessageTypeError
 import v2.models.errors.FailedToValidateError.OtherError
 import v2.models.errors.FailedToValidateError.SchemaFailedToValidateError
+import v2.models.errors.PersistenceError
 import v2.models.request.MessageType
+import v2.services.DeparturesPersistenceService
 import v2.services.ValidationService
 
 import scala.concurrent.Future
@@ -64,7 +67,8 @@ class V2DeparturesControllerImpl @Inject() (
   temporaryFileCreator: TemporaryFileCreator,
   authActionNewEnrolmentOnly: AuthNewEnrolmentOnlyAction,
   validationService: ValidationService,
-  messageSizeAction: MessageSizeAction
+  departuresPersistenceService: DeparturesPersistenceService,
+  messageSizeAction: MessageSizeAction[AuthenticatedRequest]
 )(implicit val materializer: Materializer)
     extends BaseController
     with V2DeparturesController
@@ -104,6 +108,10 @@ class V2DeparturesControllerImpl @Inject() (
     case SchemaFailedToValidateError(validationErrors) => BaseError.schemaValidationError(validationErrors = validationErrors)
   }
 
+  def translatePersistenceError(persistenceError: PersistenceError): BaseError = persistenceError match {
+    case err: PersistenceError.OtherError => BaseError.internalServiceError(cause = err.thr)
+  }
+
   def submitDeclaration(): Action[Source[ByteString, _]] =
     (authActionNewEnrolmentOnly andThen messageSizeAction).async(streamFromMemory) {
       implicit request =>
@@ -112,12 +120,18 @@ class V2DeparturesControllerImpl @Inject() (
         implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
         withTemporaryFile {
           (temporaryFile, source) =>
-            validationService
-              .validateXML(MessageType.DepartureDeclaration, source)
-              .leftMap(translateValidationError)
+            for {
+              _                 <- validationService
+                                    .validateXML(MessageType.DepartureDeclaration, source)
+                                    .leftMap(translateValidationError)
+              fileSource = FileIO.fromPath(temporaryFile)
+              declarationResult <- departuresPersistenceService
+                                    .saveDeclaration(request.eoriNumber, fileSource)
+                                    .leftMap(translatePersistenceError)
+            } yield declarationResult
         }.fold[Result](
           baseError => Status(baseError.code.statusCode)(Json.toJson(baseError)),
-          _ => Accepted
+          result => Accepted(Json.toJson(result.movementId)) // TODO: do we want the message ID too?
         )
 
     }
--- a/app/v2/controllers/actions/MessageSizeAction.scala
+++ b/app/v2/controllers/actions/MessageSizeAction.scala
@@ -31,23 +31,23 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.util.Try
 
-class MessageSizeAction @Inject() (config: AppConfig)(implicit val executionContext: ExecutionContext) extends ActionRefiner[Request, Request] {
+class MessageSizeAction[R[_] <: Request[_]] @Inject() (config: AppConfig)(implicit val executionContext: ExecutionContext) extends ActionRefiner[R, R] {
 
-  override protected def refine[A](request: Request[A]): Future[Either[Result, Request[A]]] =
+  override protected def refine[A](request: R[A]): Future[Either[Result, R[A]]] =
     contentLengthHeader(request) match {
       case Some((_, length)) => Future.successful(checkSize(length, request))
       case None =>
         Future.successful(Left(BadRequest(Json.toJson(BaseError.badRequestError("Missing content-length header"))))) // should never happen
     }
 
-  private def checkSize[A](length: String, request: Request[A]): Either[Result, Request[A]] =
+  private def checkSize[A](length: String, request: R[A]): Either[Result, R[A]] =
     Try(length.toInt).toOption match {
       case Some(x) if x <= limit => Right(request)
       case Some(_)               => Left(EntityTooLarge(Json.toJson(BaseError.entityTooLargeError(s"Your message size must be less than $limit bytes"))))
       case None                  => Left(BadRequest(Json.toJson(BaseError.badRequestError("Invalid content-length value"))))
     }
 
-  private def contentLengthHeader[A](request: Request[A]): Option[(String, String)] =
+  private def contentLengthHeader[A](request: R[A]): Option[(String, String)] =
     request.headers.headers.find(_._1.equalsIgnoreCase(HeaderNames.CONTENT_LENGTH))
 
   private lazy val limit = config.messageSizeLimit
--- a/app/v2/controllers/request/AuthenticatedRequest.scala
+++ b/app/v2/controllers/request/AuthenticatedRequest.scala
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2022 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package v2.controllers.request
+
+import play.api.mvc.Request
+import play.api.mvc.WrappedRequest
+import v2.models.EORINumber
+
+case class AuthenticatedRequest[A](eoriNumber: EORINumber, request: Request[A]) extends WrappedRequest[A](request)
--- a/app/v2/services/DeparturesPersistenceService.scala
+++ b/app/v2/services/DeparturesPersistenceService.scala
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2022 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package v2.services
+
+import akka.stream.scaladsl.Source
+import akka.util.ByteString
+import cats.data.EitherT
+import com.google.inject.ImplementedBy
+import com.google.inject.Singleton
+import v2.models.EORINumber
+import v2.models.errors.PersistenceError
+import v2.models.responses.DeclarationResponse
+
+import scala.concurrent.Future
+
+@ImplementedBy(classOf[DeparturesPersistenceServiceImpl])
+trait DeparturesPersistenceService {
+
+  def saveDeclaration(eori: EORINumber, source: Source[ByteString, _]): EitherT[Future, PersistenceError, DeclarationResponse]
+
+}
+
+@Singleton
+class DeparturesPersistenceServiceImpl extends DeparturesPersistenceService {
+  override def saveDeclaration(eori: EORINumber, source: Source[ByteString, _]): EitherT[Future, PersistenceError, DeclarationResponse] = ???
+}
--- a/test/v2/controllers/actions/AuthNewEnrolmentOnlyActionSpec.scala
+++ b/test/v2/controllers/actions/AuthNewEnrolmentOnlyActionSpec.scala
@@ -42,7 +42,7 @@ class AuthNewEnrolmentOnlyActionSpec extends AnyFreeSpec with Matchers with Mock
 
     def get(): Action[AnyContent] = authAction {
       authedRequest =>
-        Ok(authedRequest.eori)
+        Ok(authedRequest.eoriNumber.value)
     }
   }
 
--- a/app/v2/controllers/actions/AuthNewEnrolmentOnlyAction.scala
+++ b/app/v2/controllers/actions/AuthNewEnrolmentOnlyAction.scala
@@ -18,7 +18,7 @@ package v2.controllers.actions
 
 import com.google.inject.Inject
 import config.Constants._
-import controllers.actions.AuthRequest
+import v2.controllers.request.AuthenticatedRequest
 import play.api.Logging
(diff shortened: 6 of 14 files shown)


app/v2/controllers/V2DeparturesController.scala
```diff
@@ -112,12 +120,20 @@ class V2DeparturesControllerImpl @Inject() (
...
+              fileSource = FileIO.fromPath(temporaryFile)
+
+              declarationResult <- departuresPersistenceService
+                .saveDeclaration(request.eoriNumber, fileSource)
+                .leftMap(translatePersistenceError)
+            } yield declarationResult
         }.fold[Result](
           baseError => Status(baseError.code.statusCode)(Json.toJson(baseError)),
-          _ => Accepted
+          result => Accepted(Json.obj("movementId" -> result.movementId)) // TODO: do we want the message ID too? Create case object if necessary
```
drnaylor-oc, line 136: What do we want return to the traders? A proper case class will be created for this, just need to decide whether we want to just return the movement id, or include the message id too. I'm leaning towards both.

drnaylor-oc, line 136: What do we want return to the traders? A proper case class will be created for this, just need to decide whether we want to just return the movement id, or include the message id too. I'm leaning towards both.

app/v2/controllers/actions/MessageSizeAction.scala
```diff
@@ -31,23 +31,23 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.util.Try
 
-class MessageSizeAction @Inject() (config: AppConfig)(implicit val executionContext: ExecutionContext) extends ActionRefiner[Request, Request] {
+class MessageSizeAction[R[_] <: Request[_]] @Inject() (config: AppConfig)(implicit val executionContext: ExecutionContext) extends ActionRefiner[R, R] {
```
drnaylor-oc, line 34: This change is to simply allow us to pass through the type of request we have, so we can use the authenticated request if this is later in the chain.

drnaylor-oc, line 34: This change is to simply allow us to pass through the type of request we have, so we can use the authenticated request if this is later in the chain.

app/v2/controllers/request/AuthenticatedRequest.scala
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package v2.controllers.request
+
+import play.api.mvc.Request
+import play.api.mvc.WrappedRequest
+import v2.models.EORINumber
+
+case class AuthenticatedRequest[A](eoriNumber: EORINumber, request: Request[A]) extends WrappedRequest[A](request)
```
drnaylor-oc, line 23: Created because it includes the EORINumber as a value object, we could consider backporting EORINumber to phase 4 instead and just adjusting that auth request.

drnaylor-oc, line 23: Created because it includes the EORINumber as a value object, we could consider backporting EORINumber to phase 4 instead and just adjusting that auth request.

app/v2/services/DeparturesPersistenceService.scala
```diff
@@ -0,0 +1,61 @@
...
+import v2.models.EORINumber
+import v2.models.errors.PersistenceError
+import v2.models.responses.DeclarationResponse
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.control.NonFatal
+
+@ImplementedBy(classOf[DeparturesPersistenceServiceImpl])
+trait DeparturesPersistenceService {
```
drnaylor-oc, line 36: I'm not sure I like this name, it is calling `transit-movements` which is our persistence microservice, but it infers persistence here. Anyone got any better ideas?

drnaylor-oc, line 36: I'm not sure I like this name, it is calling `transit-movements` which is our persistence microservice, but it infers persistence here. Anyone got any better ideas?

app/v2/controllers/V2DeparturesController.scala
drnaylor-oc: Created a Hateoas response which contains the message and movement ID

app/v2/connectors/V2BaseConnector.scala
```diff
@@ -18,11 +18,17 @@ package v2.connectors
...
 trait V2BaseConnector extends HttpErrorFunctions {
 
   protected def validationRoute(messageType: MessageType): UrlPath =
-    UrlPath.parse(s"/transit-movements-validator/messages/${messageType.code}/validation") //transit-movements-validator/messages/IE015/validation
+    UrlPath.parse(s"/transit-movements-validator/messages/${messageType.code}/validation")
+
+  protected def movementsBaseRoute: String = "/transit-movements"
+
+  protected def movementsPostDeperatureDeclaration(eoriNumber: EORINumber): UrlPath =
+    UrlPath.parse(s"$movementsBaseRoute/movements/${eoriNumber.value}/departures/") // TODO: What is RESTful here?
```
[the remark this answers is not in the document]
drnaylor-oc: `POST /traders/:eori/movements/departures` - there is the `transit-movements` prefix for now, but this may get removed in the target microservice.

app/v2/models/responses/hateoas/HateoasDeparatureDeclarationResponse.scala
```diff
@@ -0,0 +1,83 @@
...
+object HateoasDepartureMessagePostResponse {
+
+  // in this case, we want to force a different conversion then message type would give.
+  implicit private val messageTypeWrites: Writes[MessageType] = Writes(
+    t => JsString(t.code)
+  )
+
+  implicit val hateoasMessagePostResponseWrites = (
+    (__ \ "_links").write[HateoasDepartureLinks] and
+      (__ \ "movementId").write[MovementId] and
```
drnaylor-oc, line 61: This is departure ID in phase 4, I called it movement ID because I was attempting to unify some of the models at first but figured it was a fools errand at this stage when we have nothing to unify. However, it matches what we call the movement ID.

Do we want to make this a departure ID instead, simply to minimise changes between p4 and p5?

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -0,0 +1,58 @@
...
+import v2.models.errors.PersistenceError
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+trait ErrorTranslator {
+
+  implicit class ErrorConverter[E, A](value: EitherT[Future, E, A]) {
+
+    def asBaseError(implicit c: Converter[E], ec: ExecutionContext): EitherT[Future, BaseError, A] =
```
drnaylor-oc, line 34: needs a better name, probably convertError

app/v2/connectors/PersistenceConnector.scala
```diff
@@ -0,0 +1,90 @@
...
+trait PersistenceConnector {
+
+  def sendDepartureDeclaration(eori: EORINumber, source: Source[ByteString, _])(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): Future[DeclarationResponse]
+
+}
+
+@Singleton
```
RobertLaverick, line 52: I'm concerned that I've seen this a lot, what is the expensive resource we're worried about here? Or is there another reason?

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -64,13 +63,15 @@ class V2DeparturesControllerImpl @Inject() (
   temporaryFileCreator: TemporaryFileCreator,
   authActionNewEnrolmentOnly: AuthNewEnrolmentOnlyAction,
   validationService: ValidationService,
-  messageSizeAction: MessageSizeAction
+  departuresPersistenceService: DeparturesPersistenceService,
```
RobertLaverick, line 66: Personally I think the addition of Persistence is a little redundant, IMO either a service that is responsible for Persistence (in which case, it could just be a connector) or a Departures Service.

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  implicit class ErrorConverter[E, A](value: EitherT[Future, E, A]) {
+
+    def convertError(implicit c: Converter[E], ec: ExecutionContext): EitherT[Future, BaseError, A] =
+      value.leftMap(c.convert)
+  }
+
+  trait Converter[E] {
+    def convert(input: E): BaseError
+  }
```
RobertLaverick, line 40: I like this, but oddly enough *THIS* is where I think we're starting to require an enforcement of a base error type that this can all depend on. Base Error has always been more of a presentation-layer approach to errors which is fine, but this would be stricter if it enforced a conversion from EitherT[Future, NewBaseError, A] with convertError being EitherT[Future, NewBaseError, A] and convert can then be def convert(input: NewBaseError): PresentationError - which (I think) at least has more accurate nomenclature

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -112,12 +107,16 @@ class V2DeparturesControllerImpl @Inject() (
...
-            validationService
-              .validateXML(MessageType.DepartureDeclaration, source)
-              .leftMap(translateValidationError)
+            for {
+              _ <- validationService.validateXML(MessageType.DepartureDeclaration, source).convertError
+
+              fileSource = FileIO.fromPath(temporaryFile)
+
+              declarationResult <- departuresPersistenceService.saveDeclaration(request.eoriNumber, fileSource).convertError
+            } yield declarationResult
```
RobertLaverick, line 116: The use of .convertError here reads a little code-smelly to me so I did an investigation as to why (though not a fix)

I think this is a consequence of BaseError and withTemporaryFile being entangled. Since BaseError is being used as a PresentationError, it effectively forces you to consider the presentation of the error message as a result of using the temporary file, which I'm not entirely sure is a great idea. I think we need a longer look as to how we're structuring these error types to tidy them up, and ensure that the code we're writing is actually keep seperate concerns and not tying our hands unhelpfully (I think that might be the case with withTemporaryFile atm, because of what BaseError is responsible for)

app/v2/controllers/actions/MessageSizeAction.scala
```diff
@@ -31,23 +31,23 @@ import scala.concurrent.ExecutionContext
...
-  override protected def refine[A](request: Request[A]): Future[Either[Result, Request[A]]] =
+  override protected def refine[A](request: R[A]): Future[Either[Result, R[A]]] =
     contentLengthHeader(request) match {
       case Some((_, length)) => Future.successful(checkSize(length, request))
       case None =>
         Future.successful(Left(BadRequest(Json.toJson(BaseError.badRequestError("Missing content-length header"))))) // should never happen
     }
 
-  private def checkSize[A](length: String, request: Request[A]): Either[Result, Request[A]] =
+  private def checkSize[A](length: String, request: R[A]): Either[Result, R[A]] =
```
RobertLaverick, line 43: I must have missed this before but the toOption on the Try isn't needed. You can pattern match on Success and Failure

`  private def checkSize[A](length: String, request: R[A]): Either[Result, R[A]] =
    Try(length.toInt) match {
      case Success(x) if x <= limit => Right(request)
      case Success(_)               => Left(EntityTooLarge(Json.toJson(BaseError.entityTooLargeError(s"Your message size must be less than $limit bytes"))))
      case Failure(_)               => Left(BadRequest(Json.toJson(BaseError.badRequestError("Invalid content-length value"))))
    }`

app/v2/models/responses/hateoas/HateoasDeparatureDeclarationResponse.scala
```diff
@@ -0,0 +1,83 @@
...
+import play.api.libs.json.Json
+import play.api.libs.json.OWrites
+import play.api.libs.json.Writes
+import play.api.libs.json.__
+import utils.CallOps.CallOps
+import v2.models.MessageId
+import v2.models.MovementId
+import v2.models.request.MessageType
+
+object HateoasDepartureLinks {
```
RobertLaverick, line 33: Can you outline what benefit this approach gives us over the simpler approach we took in v1?

app/v2/services/DeparturesPersistenceService.scala
RobertLaverick: Oddly enough, yeah - one or the other. (See earlier)

app/v2/connectors/PersistenceConnector.scala
```diff
@@ -0,0 +1,90 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+@ImplementedBy(classOf[PersistenceConnectorImpl])
+trait PersistenceConnector {
+
+  def sendDepartureDeclaration(eori: EORINumber, source: Source[ByteString, _])(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): Future[DeclarationResponse]
```
RobertLaverick, line 48: Considering we're using RESTful APIs this seems a very RPC way of thinking to the request. Not wrong, just not sure if this is just "named because its the scope we're worried about right now"

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -0,0 +1,191 @@
...
+          .willReturn(
+            aResponse().withStatus(OK).withBody(Json.stringify(Json.toJson(okResult)))
+          )
+      )
+
+      implicit val hc: HeaderCarrier = HeaderCarrier(extraHeaders = Seq(HeaderNames.ACCEPT -> ContentTypes.JSON))
+
+      val source = Source.single(ByteString(<test></test>.mkString, StandardCharsets.UTF_8))
+
+      whenReady(persistenceConnector.sendDepartureDeclaration(eoriNumber, source)) {
```
RobertLaverick, line 86: 😄

test/v2/controllers/ErrorTranslatorSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  "ErrorConverter#convertError" - {
+    "for a success returns the same right" in {
+      val input = Right[FailedToValidateError, Unit](()).toEitherT[Future]
+      whenReady(input.convertError.value) {
+        _ mustBe Right(())
+      }
+    }
+
+    "for an error returns the same right" in {
```
RobertLaverick, line 59: for an error returns the same **right**?

test/v2/controllers/actions/AuthNewEnrolmentOnlyActionSpec.scala
```diff
@@ -42,7 +42,7 @@ class AuthNewEnrolmentOnlyActionSpec extends AnyFreeSpec with Matchers with Mock
 
     def get(): Action[AnyContent] = authAction {
       authedRequest =>
-        Ok(authedRequest.eori)
+        Ok(authedRequest.eoriNumber.value)
```
RobertLaverick, line 45: honestly, I'm all for an implicit or an extension so we can just give these anything that extends AnyVal. Or at least the party of me that finds that habit of using AnyVals annoying

RobertLaverick: (please keep using AnyVals, just to be clear)

test/v2/services/DeparturesPersistenceServiceSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class DeparturesPersistenceServiceSpec extends AnyFreeSpec with Matchers with OptionValues with ScalaFutures with MockitoSugar {
+
+  implicit val hc: HeaderCarrier = HeaderCarrier()
+
+  "Submitting a Departure Declaration" - {
+
+    val ValidRequest: Source[ByteString, NotUsed]   = Source.single(ByteString(<schemaValid></schemaValid>.mkString, StandardCharsets.UTF_8))
+    val InvalidRequest: Source[ByteString, NotUsed] = Source.single(ByteString(<schemaInvalid></schemaInvalid>.mkString, StandardCharsets.UTF_8))
```
RobertLaverick, line 52: vals typically start with a lowercase character

RobertLaverick (changes requested):
The core plumbing of "send message to transit movements" I've not got many issues with. There is a lot of weird hand-tying as a result of BaseError that seems to be making the code awkward, and with some naming conventions, confusing. Whilst we're introducing some clever code to avoid those issues, I'm not sure if we're keeping the code all that transparent for a new developer if they had to pick it up.

app/v2/models/responses/hateoas/HateoasDeparatureDeclarationResponse.scala
drnaylor-oc: Discussed in voice chat - going for the v1 approach

app/v2/controllers/V2DeparturesController.scala
drnaylor-oc: > I think this is a consequence of BaseError and withTemporaryFile being entangled.

It's not, `withTemporaryFile` just conformed to the type required by the for comprhension - and I've made that detachment.

The original issue still stands. The problem is that, ideally, I wanted to use such a structure to make the flatmap chain with `EitherT`s quite simple -- though you may recall there were leftMaps on each call. The problem is we need to be able to unify the various service errors in some way - this basically does the conversion into a presentation error so that the left types of the EitherTs remain consistent. Ultimately, we need to do the type unification somewhere. I imagine that's why David just had his `TransitMovementError` class that basically took the role of _all_ error handling.

If we were to introduce the notion of a base service error type, such as `ServiceError`, that's fine, but we then have to make all services return `EitherT[Future, ServiceError, _]` thanks to generic invarience - and I'd argue that's just as unhelpful as what we're doing here because the service layer is now being dictated by what the presentation layer requires, where here, the presentation layer concern is done in this presentation layer.

The reality is that we could just be absolutely overdoing this. These are microservices - are we overarchitecting to the point where we're just tying ourselves in knots trying to make things look nice and enforce a very clean separation? Something has to give. We just need to decide what.

app/v2/controllers/TemporaryFiles.scala
```diff
@@ -0,0 +1,80 @@
...
+import scala.concurrent.Future
+import scala.util.Try
+
+trait TemporaryFiles {
+  val temporaryFileCreator: TemporaryFileCreator
+
+  private def deleteFile(temporaryFile: Files.TemporaryFile): Future[Unit] = {
+    temporaryFile.delete()
+    Future.successful(())
+  }
```
RobertLaverick, line 44: does this need to exist since you only call this once?

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -0,0 +1,191 @@
...
+  lazy val wsclient: WSClient   = app.injector.instanceOf[WSClient]
+  lazy val appConfig: AppConfig = app.injector.instanceOf[AppConfig]
+
+  lazy val persistenceConnector: PersistenceConnectorImpl = new PersistenceConnectorImpl(wsclient, appConfig, new TestMetrics())
+  implicit lazy val ec: ExecutionContext                  = app.materializer.executionContext
+
+  "POST /traders/:eori/message/departures" - {
+
+    lazy val okResult   = DeclarationResponse(MovementId("123"), MessageId("456"))
+    lazy val eoriNumber = EORINumber("ABC123")
```
RobertLaverick, line 67: not required for this to be merged, but if you wanted to do a touch-up on this, you could take a look at Arbitrary - https://www.scala-exercises.org/scalacheck/arbitrary. Not that this needs to have multiple iterations, but randomly generated values are good

RobertLaverick (approved):

PR #271, opened 2022-07-08
Accept departure declaration endpoint to accept json

liandra-giacomine:


app/v2/controllers/V2DeparturesController.scala
```diff
@@ -70,26 +75,56 @@ class V2DeparturesControllerImpl @Inject() (
...
-              declarationResult <- departuresService.saveDeclaration(request.eoriNumber, fileSource).asPresentation
-              _ <- routerService
-                .send(MessageType.DepartureDeclaration, request.eoriNumber, declarationResult.departureId, declarationResult.messageId, fileSource)
-                .asPresentation
-            } yield declarationResult).fold[Result](
-              presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
-              result => Accepted(HateoasDepartureDeclarationResponse(result.departureId))
-            )
-        }.toResult
+        request.headers.get("Content-Type") match {
```
drnaylor-oc, line 78: `HeaderNames.CONTENT_TYPE`

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -70,26 +75,56 @@ class V2DeparturesControllerImpl @Inject() (
...
+          case Some("application/xml")  => submitDeclarationXML
+          case Some("application/json") => submitDeclarationJSON
```
drnaylor-oc, line 80: I think play has constants that we should use here (I can't remember what the class is, I know there is `ContentTypes.JSON` and `ContentTypes.XML` - though the latter is a function and may not play well with pattern matching) 

That said, we also need to think about what we are going to instruct the traders to do. Are we going to enforce these two specific content types, or are we going to enable a range of things to be sent? Either way, we need to make it clear on the documentation what content types are accepted.

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -70,26 +75,56 @@ class V2DeparturesControllerImpl @Inject() (
...
 
+  def submitDeclarationJSON(implicit request: AuthenticatedRequest[Source[ByteString, _]]): Future[Result] = {
+    request.body.to(Sink.ignore).run()
+    Future.successful(Accepted)
+  }
+
+  def submitDeclarationXML(implicit request: AuthenticatedRequest[Source[ByteString, _]]): Future[Result] =
+    withTemporaryFile {
+      (temporaryFile, source) =>
+        logger.info("Version 2 of endpoint has been called")
```
drnaylor-oc, line 94: We probably don't want to keep this logging message long term, but for now, perhaps we should either change this to say "XML", or we should move this to the `submitDeclaration` method.

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -70,26 +75,56 @@ class V2DeparturesControllerImpl @Inject() (
...
+          _ <- routerService
+            .send(MessageType.DepartureDeclaration, request.eoriNumber, declarationResult.departureId, declarationResult.messageId, fileSource)
+            .asPresentation
+        } yield declarationResult).fold[Result](
+          presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
+          result => Accepted(HateoasDepartureDeclarationResponse(result.departureId))
+        )
+    }.toResult
+
+  def handleInvalidContentType(contentType: Option[String])(implicit request: AuthenticatedRequest[Source[ByteString, _]]): Future[Result] = {
```
drnaylor-oc, line 113: Should we move this into its own trait (along with a method that does the selection in a more generic manner?)

drnaylor-oc (commented):
Some musings.

derekspalmer (approved):
LGTM

app/v2/controllers/V2DeparturesController.scala
ardwalker: import play.api.http.ContentTypes

Looks like XML is only a function in that it wraps mimetype in a charset, think it should be ok

liandra-giacomine: @drnaylor-oc @ardwalker I used play.api.http.MimeTypes

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -70,26 +77,56 @@ class V2DeparturesControllerImpl @Inject() (
...
-                .send(MessageType.DepartureDeclaration, request.eoriNumber, declarationResult.departureId, declarationResult.messageId, fileSource)
-                .asPresentation
-            } yield declarationResult).fold[Result](
-              presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
-              result => Accepted(HateoasDepartureDeclarationResponse(result.departureId))
-            )
-        }.toResult
+        logger.info("Version 2 of endpoint has been called")
+
+        request.headers.get(HeaderNames.CONTENT_TYPE) match {
```
ardwalker, line 82: Not sure about this, but could there be an issue with case sensitivity?

drnaylor-oc: No, just looked, Play uses a `CaseInsensitiveOrdered` `TreeMap`, so everything is case insensitive here.

drnaylor-oc: (for the curious, see `play.api.mvc.Headers#toMap`)

app/v2/controllers/V2DeparturesController.scala
drnaylor-oc: Just to add, in clarification, we accepted we're only going to use `application/xml` and `application/json`. We should document this, and alongside that, document we only accept UTF-8 (as charsets can be specified in content types, and we want to explicity disallow that).

app/v2/controllers/V2DeparturesController.scala
ardwalker: Cool, thanks @drnaylor-oc

drnaylor-oc (approved):
Looks fine to me (tests haven't been removed as a cursory look at the diffs suggest, so I'm happy with that!)

PR #276, opened 2022-07-22
Validate and convert departure declaration JSON

liandra-giacomine:


app/v2/connectors/ConversionConnector.scala
```diff
@@ -0,0 +1,84 @@
...
+          .execute[HttpResponse]
+          .flatMap {
+            response =>
+              response.status match {
+                case OK =>
+                  Future.successful(response.bodyAsSource)
+                case _ =>
+                  Future.failed(UpstreamErrorResponse(response.body, response.status))
+              }
+          }
```
drnaylor-oc, line 82: This should use `stream` rather than `execute`.

Note that if we use stream, `response.body` will cause an exception -- so we should do something like I did in the auditing PR: https://github.com/hmrc/transit-movements-auditing/pull/6/files#diff-26f8b2b1726710997b8489ca5555689201073a7f278846e35aac09afae100135R61-R76

app/v2/services/ValidationService.scala
```diff
@@ -68,4 +73,25 @@ class ValidationServiceImpl @Inject() (validationConnector: ValidationConnector)
...
+    EitherT(
+      validationConnector
+        .postJson(messageType, source)
+        .map {
+          case None           => Right(())
+          case Some(response) => Left(FailedToValidateError.JsonSchemaFailedToValidateError(response.validationErrors))
+        }
+        .recover {
+          // A bad request might be returned if the stream doesn't contain XML/JSON, in which case, we need to return a bad request.
+          case UpstreamErrorResponse.Upstream4xxResponse(response) if response.statusCode == BAD_REQUEST =>
```
drnaylor-oc, line 89: They're deprecating this form. You can match on (from memory)

```scala
case UpstreamErrorResponse(_, BAD_REQUEST, _, _) => Left(....)
```

conf/application.conf
```diff
@@ -218,6 +218,12 @@ microservice {
       host = "localhost"
       port = 9498
     }
+
+    transit-movements-converter {
```
drnaylor-oc, line 222: As a note: check to see if this is in app-config-base, if not, it's worth doing it now.

app/v2/connectors/ConversionConnector.scala
```diff
@@ -70,14 +72,20 @@ class ConversionConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
           .addHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.JSON)
           .addHeaders(HeaderNames.ACCEPT -> MimeTypes.XML)
           .withBody(xmlStream)
```
derekspalmer, line 74: can we rename xmlStream to jsonStream?

derekspalmer (dismissed):
LGTM

app/v2/connectors/ConversionConnector.scala
liandra-giacomine: The input is an xmlStream, then after the conversion it will return a jsonStream

drnaylor-oc: No, it is actually the other way around!

liandra-giacomine: Oh yeah

derekspalmer (approved):

PR #277, opened 2022-07-28
CTCP-977: Remove SourceParallelisation(Spec), move auditing to after validation

drnaylor-oc:
I didn't realise that setup of the source parallelisation code would be async, and on faster machines would cause race conditions whereby the audit or persistence service wouldn't get the full payload. Because we want correctness over speed, I've removed the parallelisation, and right now it'd probably be a lot of effort to put it back in, thanks to Play WS needing sources for streaming.

As a result, the problem wasn't intermittent tests but intermittent runtime - far worse. However, the unit tests have been refactored slightly anyway because it does make them more readable.


drnaylor-oc:
(just going to run regression tests on this, but it should be fine)

drnaylor-oc:
(they work fine)

PR #282, opened 2022-08-03
CTCP-773: Add endpoint to get a message for a trader

drnaylor-oc:
Draft for the moment. This PR updates the `GET /customs/transits/movements/departures/:departureId/messages/:messageId` endpoint to accept version 2.

This PR:

- [x] Adds binders for MessageId and MovementId to only accept 16 character hex strings in the URL
- [x] Adds routing between the two versions of the GET endpoint
- [x] Adds the logic behind the version 2 endpoint

Note that for the binding, I have ensured that the `statusCode` field has remained if we can't bind to a parameter, as that is what we currently send, along with a message. I have added the "code" field too, which will bring it inline with our other messages - due to this being an addition, I think this probably isn't going to be a breaking change in v1, but I'm open (and still considering!) to changing it there to match what Play does.

### DIFF
--- a/app/controllers/DepartureMessagesController.scala
+++ b/app/controllers/DepartureMessagesController.scala
@@ -16,13 +16,16 @@
 
 package controllers
 
-import java.time.OffsetDateTime
+import com.google.inject.ImplementedBy
+import com.google.inject.Singleton
 
+import java.time.OffsetDateTime
 import com.kenshoo.play.metrics.Metrics
 import connectors.DepartureMessageConnector
 import controllers.actions.AuthAction
 import controllers.actions.ValidateAcceptJsonHeaderAction
 import controllers.actions.ValidateDepartureMessageAction
+
 import javax.inject.Inject
 import metrics.HasActionMetrics
 import metrics.MetricsKeys
@@ -46,6 +49,14 @@ import scala.concurrent.ExecutionContext
 import scala.xml.NodeSeq
 import controllers.actions.AnalyseMessageActionProvider
 
+@ImplementedBy(classOf[DepartureMessagesController])
+trait V1DepartureMessagesController {
+
+  def getDepartureMessage(departureId: DepartureId, messageId: MessageId): Action[AnyContent]
+
+}
+
+@Singleton
 class DepartureMessagesController @Inject() (
   cc: ControllerComponents,
   authAction: AuthAction,
@@ -58,7 +69,8 @@ class DepartureMessagesController @Inject() (
     extends BackendController(cc)
     with HasActionMetrics
     with HttpErrorFunctions
-    with ResponseHelper {
+    with ResponseHelper
+    with V1DepartureMessagesController {
 
   import MetricsKeys.Endpoints._
 
@@ -86,7 +98,7 @@ class DepartureMessagesController @Inject() (
                                 request.body
                               )
                             )
-                          ).withHeaders(LOCATION -> routes.DepartureMessagesController.getDepartureMessage(departureId, messageId).urlWithContext)
+                          ).withHeaders(LOCATION -> routing.routes.DeparturesRouter.getMessage(departureId.toString, messageId.toString).urlWithContext)
                         case None =>
                           InternalServerError
                       }
--- a/app/routing/DeparturesRouter.scala
+++ b/app/routing/DeparturesRouter.scala
@@ -22,8 +22,8 @@ import akka.util.ByteString
 import com.google.inject.Inject
 import controllers.V1DepartureMessagesController
 import controllers.V1DeparturesController
-import models.domain.DepartureId
-import models.domain.MessageId
+import models.domain.{DepartureId => V1DepartureId}
+import models.domain.{MessageId => V1MessageId}
 import play.api.mvc.Action
 import play.api.mvc.BaseController
 import play.api.mvc.ControllerComponents
@@ -31,7 +31,8 @@ import play.api.mvc.PathBindable
 import v2.controllers.V2DeparturesController
 import v2.controllers.stream.StreamingParsers
 import v2.models.Bindings._
-import v2.models.MovementId
+import v2.models.{DepartureId => V2DepartureId}
+import v2.models.{MessageId => V2MessageId}
 
 class DeparturesRouter @Inject() (
   val controllerComponents: ControllerComponents,
@@ -52,16 +53,16 @@ class DeparturesRouter @Inject() (
   def getMessage(departureId: String, messageId: String): Action[Source[ByteString, _]] = route {
     case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE) =>
       (for {
-        convertedDepartureId <- implicitly[PathBindable[MovementId]].bind("departureId", departureId)
-        convertedMovementId  <- implicitly[PathBindable[v2.models.MessageId]].bind("messageId", messageId)
+        convertedDepartureId <- implicitly[PathBindable[V2DepartureId]].bind("departureId", departureId)
+        convertedMovementId  <- implicitly[PathBindable[V2MessageId]].bind("messageId", messageId)
       } yield (convertedDepartureId, convertedMovementId)).fold(
         bindingFailureAction(_),
         converted => v2Departures.getMessage(converted._1, converted._2)
       )
     case _ =>
       (for {
-        convertedDepartureId <- implicitly[PathBindable[DepartureId]].bind("departureId", departureId)
-        convertedMovementId  <- implicitly[PathBindable[MessageId]].bind("messageId", messageId)
+        convertedDepartureId <- implicitly[PathBindable[V1DepartureId]].bind("departureId", departureId)
+        convertedMovementId  <- implicitly[PathBindable[V1MessageId]].bind("messageId", messageId)
       } yield (convertedDepartureId, convertedMovementId)).fold(
         bindingFailureAction(_),
         converted => v1DepartureMessages.getDepartureMessage(converted._1, converted._2)
--- a/app/routing/DeparturesRouter.scala
+++ b/app/routing/DeparturesRouter.scala
@@ -54,16 +54,16 @@ class DeparturesRouter @Inject() (
     case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE) =>
       (for {
         convertedDepartureId <- implicitly[PathBindable[V2DepartureId]].bind("departureId", departureId)
-        convertedMovementId  <- implicitly[PathBindable[V2MessageId]].bind("messageId", messageId)
-      } yield (convertedDepartureId, convertedMovementId)).fold(
+        convertedMessageId   <- implicitly[PathBindable[V2MessageId]].bind("messageId", messageId)
+      } yield (convertedDepartureId, convertedMessageId)).fold(
         bindingFailureAction(_),
         converted => v2Departures.getMessage(converted._1, converted._2)
       )
     case _ =>
       (for {
         convertedDepartureId <- implicitly[PathBindable[V1DepartureId]].bind("departureId", departureId)
-        convertedMovementId  <- implicitly[PathBindable[V1MessageId]].bind("messageId", messageId)
-      } yield (convertedDepartureId, convertedMovementId)).fold(
+        convertedMessageId   <- implicitly[PathBindable[V1MessageId]].bind("messageId", messageId)
+      } yield (convertedDepartureId, convertedMessageId)).fold(
         bindingFailureAction(_),
         converted => v1DepartureMessages.getDepartureMessage(converted._1, converted._2)
       )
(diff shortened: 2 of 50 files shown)


app/routing/DeparturesRouter.scala
```diff
@@ -20,22 +20,54 @@ import akka.stream.Materializer
...
-    case Some("application/vnd.hmrc.2.0+json") => v2Departures.submitDeclaration()
-    case _                                     => v1Departures.submitDeclaration()
+    case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE) => v2Departures.submitDeclaration()
+    case _                                                    => v1Departures.submitDeclaration()
+  }
+
+  def getMessage(departureId: String, messageId: String): Action[Source[ByteString, _]] = route {
+    case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE) =>
+      (for {
+        convertedDepartureId <- implicitly[PathBindable[MovementId]].bind("departureId", departureId)
```
liandra-giacomine, line 56: I'm a bit confused about the use of movementId vs messageId vs departureId...Here we bind departureId to MovementId but we have the convertedMovementId which binds the messageId? Sorry about the ignorance, just need to understand what the link is :)

conf/app.routes
```diff
@@ -19,6 +19,6 @@ GET        /movements/departures/:departureId                       controllers.
 POST       /movements/departures/:departureId/messages              controllers.DepartureMessagesController.sendMessageDownstream(departureId: DepartureId)
 GET        /movements/departures/:departureId/messages              controllers.DepartureMessagesController.getDepartureMessages(departureId: DepartureId, receivedSince: Option[OffsetDateTime] ?= None)
 
-GET        /movements/departures/:departureId/messages/:messageId   controllers.DepartureMessagesController.getDepartureMessage(departureId: DepartureId, messageId: MessageId)
+GET        /movements/departures/:departureId/messages/:messageId   routing.DeparturesRouter.getMessage(departureId: String, messageId: String)
```
liandra-giacomine, line 22: Just wondering why you chose to declare departureId/messageId as strings rather than case classes extending anyval?

app/routing/DeparturesRouter.scala
drnaylor-oc: The crux of this is that you're looking at the routing layer.

In v2, we've not really made a distinction between a departure and what will be an arrival id (both will be `MovementId`), because they're both short UUIDs and I'd imagine that some of the other services may be completely agnostic to them - indeed, it's entirely possible that transit-movements won't care about the distinction, so our connector won't either. 

As a result, `departureId` is just the name of the parameter that we're using in the URL, it will bind to a `MovementId` - this is mostly because we're trying to keep the URLs the same as v1 where possible.

We have this slightly complicated mess of binding because of a significant difference between v1 and v2 (and which is why I'm simply binding to a string here and thus in the routes), v1 uses integers as the departure/message IDs, v2 uses hex strings for the movement/message IDs. Because we want to use the same endpoint, we can't bind the parameters to an int or a hex string model, so instead I effectively say "I'm going to take over the binding phase, give me a string and I'll do the appropriate conversions".

conf/app.routes
drnaylor-oc: See my other comment - basically because I do the binding myself once I've decided which version to send it on to.

app/routing/DeparturesRouter.scala
drnaylor-oc: I've just been looking around the code and what we're doing and it looks like transit-movements will make a distinction, right down to the fact the URL has "departure" in there.

So that's a good question we need to ask - is `MovementId` something we should keep, or should be split it to `DepartureId` and later on `ArrivalId`?

app/routing/DeparturesRouter.scala
```diff
@@ -52,16 +53,16 @@ class DeparturesRouter @Inject() (
   def getMessage(departureId: String, messageId: String): Action[Source[ByteString, _]] = route {
     case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE) =>
       (for {
-        convertedDepartureId <- implicitly[PathBindable[MovementId]].bind("departureId", departureId)
-        convertedMovementId  <- implicitly[PathBindable[v2.models.MessageId]].bind("messageId", messageId)
+        convertedDepartureId <- implicitly[PathBindable[V2DepartureId]].bind("departureId", departureId)
+        convertedMovementId  <- implicitly[PathBindable[V2MessageId]].bind("messageId", messageId)
```
liandra-giacomine, line 57: shouldn't counvertedMovementId be convertedMessageId?

drnaylor-oc: Very much so!

app/controllers/DepartureMessagesController.scala
```diff
@@ -86,7 +98,7 @@ class DepartureMessagesController @Inject() (
                                 request.body
                               )
                             )
-                          ).withHeaders(LOCATION -> routes.DepartureMessagesController.getDepartureMessage(departureId, messageId).urlWithContext)
+                          ).withHeaders(LOCATION -> routing.routes.DeparturesRouter.getMessage(departureId.toString, messageId.toString).urlWithContext)
```
liandra-giacomine, line 101: I think we should be using .value for case classes that extend anyval rather than toString, one just retrieves the value already in memory and the other executes a function to derive a value each time it's called?

drnaylor-oc: Truth be told, I don't know why I used toString... my bad.

drnaylor-oc: Oh, yes, I do, now I looked at it. In P4, these are interger backed values, not string. So we need to do the conversion anyway.

liandra-giacomine: But wouldn't it already be a string when you are retrieving the value from the case class?

liandra-giacomine: I get the initial conversion to set the departureId value

drnaylor-oc: No - it's an integer backed class.

drnaylor-oc: https://github.com/hmrc/common-transit-convention-traders/blob/acbc65b663825309f72783a4c27d0028cab56fd7/app/models/domain/DepartureId.scala#L21

drnaylor-oc: (didn't mean to click the resolve button!)

liandra-giacomine: We have two DepartureIds? Sorry if I'm being slow: https://github.com/hmrc/common-transit-convention-traders/pull/282/files#diff-f6c5dac1148800169cd955c75ed3164515e472a20629e842742b25aefb56f8a7

liandra-giacomine: Also, shouldn't we make the def toString in the case class a val, so we only have to calculate the string value once?

drnaylor-oc: We do - the P4 is an integer backed one, the P5 one is a string backed one.

We can't make the def a val, because then it can't be an AnyVal.

liandra-giacomine: Gotcha, question time is over ;) you can mark this as resolved now :D

app/routing/DeparturesRouter.scala
drnaylor-oc: I was chatting with Rob, we've made it a `DepartureId` to be consistent.

app/v2/connectors/V2BaseConnector.scala
```diff
@@ -34,9 +34,12 @@ trait V2BaseConnector extends HttpErrorFunctions {
   protected def movementsPostDeperatureDeclaration(eoriNumber: EORINumber): UrlPath =
     UrlPath.parse(s"$movementsBaseRoute/traders/${eoriNumber.value}/movements/departures/")
 
+  protected def movementsGetDepartureMessage(eoriNumber: EORINumber, departureId: DepartureId, messageId: MessageId): UrlPath =
+    UrlPath.parse(s"$movementsBaseRoute/traders/${eoriNumber.value}/movements/departures/${departureId.value}/messages/${messageId.value}/")
+
```
RobertLaverick, line 39: Why isn't this using the servicesConfig object? That would allow you to source movementsBaseRoute from the appconfig.

conf/app.routes
RobertLaverick: To me that suggests we could probably do a refactor to make the Id-types work on the same basis across P4 and P5. Means we can have a common pattern now and avoid the tidy-up before handover to live when P5 drops

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -72,21 +80,21 @@ class PersistenceConnectorSpec
...
+  lazy val shortUuidGen: Arbitrary[String] = Arbitrary(Gen.long.map {
+    l: Long =>
+      f"${BigInt(l)}%016x"
+  })
 
-    lazy val shortUuidGen: Arbitrary[String] = Arbitrary(Gen.long.map {
-      l: Long =>
-        f"${BigInt(l)}%016x"
-    })
+  lazy val eoriNumberGen = Gen.alphaNumStr.map(EORINumber.apply)
```
RobertLaverick, line 88: Surely these are going to be reused elsewhere? Lets move them into an importable trait for now.

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -208,6 +216,167 @@ class PersistenceConnectorSpec
     }
   }
 
+  "GET /traders/:eori/movements/departure/:departureid/messages/:message" - {
+
+    lazy val shortUuidGen: Arbitrary[String] = Arbitrary(Gen.long.map {
+      l: Long =>
+        f"${BigInt(l)}%016x"
+    })
```
RobertLaverick, line 224: Isn't this duplication of line 75?

test/routing/DeparturesRouterSpec.scala
```diff
@@ -16,78 +16,196 @@
...
-      bind[V1DeparturesController].to[FakeV1DeparturesController],
-      bind[V2DeparturesController].to[FakeV2DeparturesController]
-    )
-    .build()
+  val sut = new DeparturesRouter(
+    stubControllerComponents(),
+    new FakeV1DeparturesController,
+    new FakeV1DepartureMessagesController,
+    new FakeV2DeparturesController
+  )
```
RobertLaverick, line 52: nice

RobertLaverick (changes requested):

conf/app.routes
drnaylor-oc: > To me that suggests we could probably do a refactor to make the Id-types work on the same basis across P4 and P5.

I'm not sure how you would want to refactor this.

* If you wanted to unify each of the Id types so it was common across V1 and V2, I'd be against that because they are fundamentally different types, and it means adding more code elsewhere to ensure correctness later.
* If you wanted to add a way to unify this handling, that I can see an argument of, but I'm wary about potentially wading into "clever code" for the sake of it.
* Is there something else you had in mind?

While we have come up with some clever tricks in this codebase, I'm not really sure we need to do much more here. Once P5 is live and P4 is dead to us, I'd imagine this router will be deleted, along with the P4 code, and the routes file will be altered to just support the V2 routes (along with the reverse routing breaks that will occur in the tests and all...). By doing so, this would just outright delete the binding selection and we can just use the right model in this routes file. I'd be happy to include the P5 only route as a comment.

app/v2/connectors/V2BaseConnector.scala
drnaylor-oc: I don't really care where this is, but I keep hearing this argument flip flop both ways - when do we use a config option vs. just doing it in the code because either way, it requires a release (and we're not going to do it on a per-environment basis, which is the only way to update a config variable without a new release).

If we make it us the `servicesConfig` object, then we need to ensure we inject it in all connectors and that it's defined here. It's not an issue, but it's what we'd need to do.

it/v2/connectors/PersistenceConnectorSpec.scala
drnaylor-oc: Yup!

it/v2/connectors/PersistenceConnectorSpec.scala
drnaylor-oc: ...yup

RobertLaverick (approved):

app/v2/connectors/RouterConnector.scala
```diff
@@ -27,25 +27,21 @@ import metrics.MetricsKeys
...
-import uk.gov.hmrc.http.HttpReads.Implicits._
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
 @ImplementedBy(classOf[RouterConnectorImpl])
 trait RouterConnector {
 
-  def post(messageType: MessageType, eoriNumber: EORINumber, movementId: MovementId, messageId: MessageId, body: Source[ByteString, _])(implicit
+  def post(messageType: MessageType, eoriNumber: EORINumber, movementId: DepartureId, messageId: MessageId, body: Source[ByteString, _])(implicit
```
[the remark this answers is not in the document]
drnaylor-oc: I'll do it in the next PR I've got going on.

PR #283, opened 2022-08-04
Ctcp 737

derekspalmer:


app/v2/models/HeaderType.scala
```diff
@@ -0,0 +1,37 @@
...
+
+object HeaderTypes {
+
+  case object jsonToXml
+      extends HeaderType(
+        (HeaderNames.CONTENT_TYPE -> MimeTypes.JSON),
+        (HeaderNames.ACCEPT       -> MimeTypes.XML)
+      )
+
+  case object xmlToJson
```
liandra-giacomine, line 32: I may be wrong, but I don't think the XML -> JSON case is required here since it will be the auditing service that will send the request to convert the XML to JSON?

app/v2/services/ConversionService.scala
```diff
@@ -36,25 +39,37 @@ import scala.util.control.NonFatal
...
     materializer: Materializer
   ): EitherT[Future, ConversionError, Source[ByteString, _]]
 
+  def jsonToXml(messageType: MessageType, source: Source[ByteString, _])(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext,
+    materializer: Materializer
+  ): EitherT[Future, ConversionError, Source[ByteString, _]] = convert(messageType, source, HeaderTypes.jsonToXml)
+
+  def xmlToJson(messageType: MessageType, source: Source[ByteString, _])(implicit
```
liandra-giacomine, line 54: Similarly, we won't be converting from XML to JSON here, I don't think

app/v2/models/HeaderType.scala
derekspalmer: Not sure - probably need to Tim or Dan to clarify!

drnaylor-oc: We might need it for the Web FE, they _might_ want the messages converted to Json if they poll our service. If so, we'll probably need to do it here.

@crystalmark might have a bit of a better idea as to how we're going to work with the web front end on this regard -- indeed, we may need to talk to them about it.

it/v2/connectors/ConversionConnectorSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+      server.stubFor(
+        post(
+          urlEqualTo(s"/transit-movements-converter/messages/${messageType.code}")
+        ).willReturn(
+          aResponse().withStatus(500).withBody("Internal service error")
+        )
+      )
+
+      val postResponse = sut.post(MessageType.DepartureDeclaration, jsonStream, jsonToXml)
+      // whenReady for some reason not working - even when changing patienceConfig!
```
drnaylor-oc, line 86: It won't work because you have a failed future, whenReady only works in the success case.

I thought `Future#map` also only worked on the success case, so I suspect that your test is passing simply because the map isn't executing so the assertation isn't firing.

You'll want to do something that checks the failure condition, you can either map to a right and recover to a left, then `whenReady` that, i.e.

```java
future.map(Right(_)).recover(Left(_))
```

or you could fail directly on `map` and pattern match on `recover`, and whenReady on that - I've done something like this elsewhere:

```java
        val r = persistenceConnector
          .getDepartureMessage(eori, departureId, messageId)
          .map(
            _ => fail("This should have failed with an UpstreamErrorResponse, but it succeeded")
          )
          .recover {
            case UpstreamErrorResponse(_, NOT_FOUND, _, _) => ()
            case thr                                       => fail(s"Expected an UpstreamErrorResponse with a 404, got $thr")
          }

        whenReady(r) {
          _ =>
        }
}
```

derekspalmer: Yes, of course. It's a failing test!

app/v2/services/ConversionService.scala
derekspalmer: See above.

ardwalker (approved):

PR #284, opened 2022-08-05
Ctcp 783

ardwalker:


crystalmark (approved):

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -73,7 +85,13 @@ class V2DeparturesControllerImpl @Inject() (
...
     with ErrorTranslator
     with TemporaryFiles
-    with ContentTypeRouting {
+    with ContentTypeRouting
+    with HasActionMetrics {
+
+  lazy val sCounter: Counter = counter(s"success-counter")
+  lazy val fCounter: Counter = counter(s"failure-counter")
+
+  lazy val gauge: Gauge[_] = registry.gauge(s"error-ratio", () => ErrorRatioGauge(sCounter, fCounter))
```
drnaylor-oc, line 94: Looping back to https://github.com/hmrc/grafana-dashboards/pull/2078/files - you're not using this (and that's perfectly fine, as if we used this metric, Grafana would average this over the instances, not the sum of the two counts). 

If we're not using this, we should remove it, along with `ErrorRatioGauge`.

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -87,19 +105,46 @@ class V2DeparturesControllerImpl @Inject() (
...
 
+  def persistAndSendToEIS(
+    src: Source[ByteString, _]
+  )(implicit hc: HeaderCarrier, request: AuthenticatedRequest[Source[ByteString, _]]): EitherT[Future, PresentationError, DeclarationResponse] =
+    withTemporaryFileA(
+      src,
+      (temporaryFile, _) => {
+        val fileSource = FileIO.fromPath(temporaryFile)
+        (for {
+          declarationResult <- departuresService.saveDeclaration(request.eoriNumber, fileSource).asPresentation
```
drnaylor-oc, line 140: Is there scope to deduplicate this with the equivalent section in `submitDeclarationXML`?

app/v2/controllers/V2DeparturesController.scala
ardwalker: Will remove, since we can get the functionality we need from Grafana

app/v2/controllers/V2DeparturesController.scala
ardwalker: Done

drnaylor-oc (approved):

PR #287, opened 2022-08-10
CTCP-783 Fixed issue where the same single use Source was being reused

ardwalker:


app/v2/controllers/ErrorTranslator.scala
```diff
@@ -42,6 +42,14 @@ trait ErrorTranslator {
     def convert(input: E): PresentationError
   }
 
+  val jsonToXmlValidationErrorConverter = new Converter[FailedToValidateError] {
+
+    def convert(validationError: FailedToValidateError): PresentationError = validationError match {
+      case XmlSchemaFailedToValidateError(_) => PresentationError.internalServiceError(cause = None) // TODO: Determine error message
```
drnaylor-oc, line 48: On the TODO: the error message is actually fine -- we want to display "Internal server error" to the user.

We may need to consider how we show the exceptions in the logs in a way that isn't going to display PII... but I think that's out of scope for here because it's a wider issue and we really just want to get this done now.

ardwalker: Agree, it's a difficult one, because we  want to know what went wrong so it can be fixed, but it's not something the end user can do anything about. Logging sounds like a good option - but are you saying we cannot just log the output because it may contain sensitive information?

drnaylor-oc: Correct - if you think about what the validation service returns, it returns where things failed. It's possible that it'll present a string and say "this failed because of this regex". We can't log that.

We may just need to log that validation failed for a given request ID. With the intention of auditing these failures, we can then turn to the audit platform to grab the payload if we need to.

Such logging wouldn't be done here though, I'd suspect it's a service layer thing.

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -42,6 +42,14 @@ trait ErrorTranslator {
     def convert(input: E): PresentationError
   }
 
+  val jsonToXmlValidationErrorConverter = new Converter[FailedToValidateError] {
```
drnaylor-oc, line 45: We should probably get a set of unit tests for this as this is a new converter. It should suffice to copy the Validation Error block in the tests:

https://github.com/hmrc/common-transit-convention-traders/blob/a71421e279960fe95f525c9910ce0c8a48c2e706/test/v2/controllers/ErrorTranslatorSpec.scala#L64-L105

and then just ensure the XML validation error test is changed to return an Internal Server Error.

drnaylor-oc (approved):

PR #289, opened 2022-08-15
CTCP-834 GET departure by movement id (FE API)

ardwalker:


drnaylor-oc:
You're missing the API definition in `resources/public/api/conf/2.0/application.raml` (look at the 1.0 raml file as a template for this endpoint).

I only remembered because I have only just remembered to do it myself!

ardwalker:
> You're missing the API definition in `resources/public/api/conf/2.0/application.raml` (look at the 1.0 raml file as a template for this endpoint).
> 
> I only remembered because I have only just remembered to do it myself!

Thanks @drnaylor-oc , updated

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -858,6 +862,86 @@ class V2DeparturesControllerSpec
 
   }
 
+  // for retrieving a single departure/movement
+  "GET  /movements/departures/:departureId" - {
```
RobertLaverick, line 866: Is it worth putting a Tech Debt ticket into the backlog to rework these tests using model generators? (Or at least to use fewer magic values)

RobertLaverick (changes requested):

app/routing/DeparturesRouter.scala
```diff
@@ -22,6 +22,7 @@ import akka.util.ByteString
 import com.google.inject.Inject
 import controllers.V1DepartureMessagesController
 import controllers.V1DeparturesController
+//import controllers.V1DepartureController
```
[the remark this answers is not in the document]
ardwalker: Removing

app/v2/models/responses/DepartureResponse.scala
```diff
@@ -0,0 +1,37 @@
...
+
+object DepartureResponse {
+  implicit lazy val departureResponseFormat = Json.format[DepartureResponse]
+}
+
+case class DepartureResponse(
+  _id: DepartureId,
+  enrollmentEORINumber: EORINumber,
+  movementEORINumber: EORINumber,
+  movementReferenceNumber: Option[MovementReferenceNumber], // optional pending MRN allocation
```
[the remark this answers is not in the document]
ardwalker: Good question - happy to remove.

test/v2/controllers/V2DeparturesControllerSpec.scala
ardwalker: Might be easier to just rename the test, so rather than using the URI, I can use the comment above it as the name of the test?

app/v2/models/responses/hateoas/HateoasDepartureResponse.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import play.api.libs.json.JsObject
+import play.api.libs.json.Json
+import v2.models.DepartureId
+import v2.models.responses.DepartureResponse
+
+object HateoasDepartureResponse {
+
+  def prefix(departureId: DepartureId) =
+    if (routing.routes.DeparturesRouter.getDeparture(departureId.value).url.startsWith("/customs/transits")) ""
```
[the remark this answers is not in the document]
drnaylor-oc: We do have it as a constant actually (and I wasn't aware of this): `Constants.Context`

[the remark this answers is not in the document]
ardwalker: Found it, I was not aware either

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -858,6 +862,85 @@ class V2DeparturesControllerSpec
 
   }
 
+  "retrieving a single departure/movement" - {
```
RobertLaverick, line 865: That wasn't what I meant, the name of the test was fine.

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -858,6 +862,85 @@ class V2DeparturesControllerSpec
...
+    "should return ok with json body of departure" in {
+      val createdTime = OffsetDateTime.now()
+      val departureResponse = DepartureResponse(
+        DepartureId("0123456789abcdef"),
+        EORINumber("GB123"),
+        EORINumber("XI456"),
+        Some(MovementReferenceNumber("312")),
```
RobertLaverick, line 872: These are the magic numbers I was referring to.

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -858,6 +862,85 @@ class V2DeparturesControllerSpec
...
+
+      status(result) mustBe OK
+      contentAsJson(result) mustBe Json.toJson(
+        HateoasDepartureResponse(
+          DepartureId("0123456789abcdef"),
+          DepartureResponse(
+            DepartureId("0123456789abcdef"),
+            EORINumber("GB123"),
+            EORINumber("XI456"),
+            Some(MovementReferenceNumber("312")),
```
RobertLaverick, line 898: As are these

ardwalker: Ok thanks for letting me know. Just so we don’t need to go back and forth, can you please clarify the scope of your ask? As I mentioned earlier, magic strings are everywhere. Have just been following other example code, so I am guessing we need to create some tech debt stories to eliminate magic strings?

Is there already/or is it worth writing a standards doc to simplify the PR process?

Thanks.

drnaylor-oc: I think his point was more that we should open a tech debt ticket to consider replacing the use of these hard coded values in the test with generators so that we're testing arbitary values rather than a specific value. You're right in the fact it's being done everywhere and it would be ludicrous to ask you to go change everything in this PR.

For example, [in some of the integration tests, we've created model generators that generate random numbers/strings each time.](https://github.com/hmrc/common-transit-convention-traders/blob/9cacbd8e085d5a79d1458ed28a792d828fe9e3dc/it/v2/connectors/PersistenceConnectorSpec.scala#L87-L115). It just heightens the chance we'll catch something weird in edge cases.

I'd definately argue that there is a case for them, but equally, I'd argue that there are times where you just need a static valued test - it all boils down to judgement and style, really. I suspect that this is one of those times where it could benefit.

If you fancy giving that a go with your stuff, great, but don't worry about doing it with anything that is outside the scope of your ticket.

ardwalker: Oh, so the ask is not about magic strings, more around not using fixed values in tests - so using dynamically generated values/property based testing? Ok, will take a look at the examples - thanks for the explanation :)

it/connectors/DepartureConnectorSpec.scala
```diff
@@ -201,7 +200,7 @@ class DepartureConnectorSpec
         LocalDateTime.now
       )
 
-      val response = HateoasResponseDeparture(departure)
```
drnaylor-oc, line 204: This seems to be in the v1/phase 4 code - did you mean to touch this? The file and class doesn't seem to be renamed (and we should keep it that way really), the only P4 stuff that should be touched are the routes file changes.

it/connectors/DepartureMessageConnectorSpec.scala
```diff
@@ -98,7 +97,7 @@ class DepartureMessageConnectorSpec
         LocalDateTime.now
       )
 
-      val response = HateoasResponseDeparture(departure)
```
drnaylor-oc, line 101: and here

resources/public/api/conf/2.0/application.raml
```diff
@@ -115,6 +115,44 @@ traits:
...
+        (annotations.scope): "common-transit-convention-traders"
+        securedBy: [ sec.oauth_2_0: { scopes: [ "common-transit-convention-traders" ] } ]
+        responses:
+          200:
+            body:
+              application/json:
+                description: JSON Payload
+                example: |
+                  {
+                    "_id": "1",
```
drnaylor-oc, line 138: Is this right? This should match the actual output somewhat - so this should be `id: /customs/transits/...`

resources/public/api/conf/2.0/application.raml
```diff
@@ -115,6 +115,44 @@ traits:
...
+                    "enrollmentEORINumber": "GB002",
+                    "movementEORINumber": "GB001",
+                    "movementReferenceNumber": "MRN001",
+                    "created": "2022-08-14T00:00:00",
+                    "updated": "2020-08-15T01:23:45",
+                    "_links": {
+                      "self": {
+                        "href": "/customs/transits/movements/departures/1"
```
drnaylor-oc, line 146: We should make the IDs match the scheme in our examples, so this should be some 16 character hex string.

app/v2/models/responses/hateoas/HateoasDepartureResponse.scala
```diff
@@ -0,0 +1,51 @@
...
+  // TODO: Update when v2 GET departure-messages is available
+  def messagesUri() = "v2-departure-messages-not-yet-implemented"
+
+  def apply(departureId: DepartureId, departureResponse: DepartureResponse): JsObject =
+    Json
+      .obj(
+        "_links" -> Json.obj(
+          "self" -> Json.obj("href" -> selfUrl(departureId), "messages" -> messagesUri())
+        ),
+        "_id"                     -> departureResponse._id,
```
drnaylor-oc, line 43: This should be ID, and the ID is the URI

it/connectors/DepartureConnectorSpec.scala
ardwalker: Well caught - I did a refactoring which affected many more files than I thought it would and forgot to change these back, thanks!

resources/public/api/conf/2.0/application.raml
ardwalker: I have been used to an ID being some form of unique identity and the URI being referred to as an HREF, but will change to make it consistent.

drnaylor-oc (approved):

PR #293, opened 2022-08-23
Ctcp 963

derekspalmer:


app/v2/models/responses/hateoas/HateoasDepartureIdsResponse.scala
```diff
@@ -0,0 +1,33 @@
...
+
+object HateoasDepartureIdsResponse extends HateoasResponse {
+
+  def apply(departureIds: Seq[DepartureId]): JsObject =
+    Json.obj(
+      "_links" -> Json.obj(
+        "self" -> Json.obj("href" -> "/customs/transits/movements/departures")
+      ),
+      "ids" -> departureIds.map(
+        id => Json.obj("href" -> departureUri(id))
```
drnaylor-oc, line 30: We should match what we're doing with message IDs here, so:

```
"departures": [
   {
      "id": "/customs/transits/movements/departures/<id>"
   }
]
```

not what you have here.

derekspalmer: Agree. Will change.

PR #294, opened 2022-08-24
CTCP-963 

RobertLaverick:
removed commented out queryparams block

### DIFF
--- a/resources/public/api/conf/2.0/application.raml
+++ b/resources/public/api/conf/2.0/application.raml
@@ -122,12 +122,12 @@ traits:
       is:
         - acceptJsonHeader
         - acceptHeaderInvalid
-#      queryParameters:
-#        updatedSince:
-#          required: false
-#          type: datetime
-#          description: This is an optional filtering parameter. It allows you to request only the Movement Departures which have been updated since the date and time you have given us.
-#          example: 2021-06-21T09:00+00:00
+      queryParameters:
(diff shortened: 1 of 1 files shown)


liandra-giacomine (approved):

resources/public/api/conf/2.0/application.raml
```diff
@@ -122,12 +122,12 @@ traits:
...
-#          type: datetime
-#          description: This is an optional filtering parameter. It allows you to request only the Movement Departures which have been updated since the date and time you have given us.
-#          example: 2021-06-21T09:00+00:00
+      queryParameters:
+        updatedSince:
+          required: false
+          type: datetime
+          description: This is an optional filtering parameter. It allows you to request only the Movement Departures which have been updated since the date and time you have given us.
+          example: 2021-06-21T09:00+00:00
         (annotations.scope): "common-transit-convention-traders"
```
drnaylor-oc, line 131: The indentation seems wrong here - should this be one level back?

(I realise you haven't touched it, but that might be why we're failing)

RobertLaverick: looked right to me, but I think it was the commenting as RAML doesn't support YAML-style comments. I'll merge and find out

drnaylor-oc: It does, I did it before.

PR #297, opened 2022-09-13
Changes related to CTCP-1092

sonisn1988:


app/v2/controllers/V2DeparturesController.scala
```diff
@@ -237,4 +238,8 @@ class V2DeparturesControllerImpl @Inject() (
             response => Ok(Json.toJson(HateoasDepartureIdsResponse(response)))
           )
     }
+
+  def sendMessageDownstream(departureId: DepartureId): Action[Source[ByteString, _]] = Action(streamFromMemory) {
+    _ => Accepted(Json.obj("version" -> 2))
```
drnaylor-oc, line 243: Though it's just a stub and this will eventually be properly implemented, we need to handle the stream properly by doing something with it. We should do the following:

```suggestion
    implicit request => 
        // we must do something with the request otherwise it'll be held in memory.
        request.body.runWith(Sink.ignore)
        Accepted(Json.obj("version" -> 2))
```

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -237,4 +238,8 @@ class V2DeparturesControllerImpl @Inject() (
             response => Ok(Json.toJson(HateoasDepartureIdsResponse(response)))
           )
     }
+
+  def sendMessageDownstream(departureId: DepartureId): Action[Source[ByteString, _]] = Action(streamFromMemory) {
```
drnaylor-oc, line 242: While this was the name given to the method in version 1 of the API, I'm not sure I like it. I think I'd prefer something like "attachMessage", rather than "sendMessageDownstream", which is only one part of what is going on here.

app/routing/DeparturesRouter.scala
```diff
@@ -101,4 +101,19 @@ class DeparturesRouter @Inject() (
     case _                                                    => v1Departures.getDeparturesForEori(updatedSince)
   }
 
+  def sendMessageDownstream(departureId: String): Action[Source[ByteString, _]] = route {
```
drnaylor-oc, line 104: A similar issue here - I'd rather name this after the version 2 action, rather than the version 1 one.

test/routing/DeparturesRouterSpec.scala
```diff
@@ -267,4 +267,50 @@ class DeparturesRouterSpec extends AnyFreeSpec with Matchers with OptionValues w
       }
     }
   }
+
+  "when submitting a movements departure" - {
```
drnaylor-oc, line 271: That's not quite what we're doing here - we're submitting a new message for an existing departure.

app/v2/controllers/V2DeparturesController.scala
sonisn1988: done

app/v2/controllers/V2DeparturesController.scala
sonisn1988: done

app/routing/DeparturesRouter.scala
sonisn1988: done

test/routing/DeparturesRouterSpec.scala
sonisn1988: done

app/controllers/DepartureMessagesController.scala
```diff
@@ -56,6 +56,8 @@ trait V1DepartureMessagesController {
 
   def getDepartureMessages(departureId: DepartureId, receivedSince: Option[OffsetDateTime]): Action[AnyContent]
 
+  def sendMessageDownstream(departureId: DepartureId): Action[NodeSeq]
```
ardwalker, line 59: you might want to rename this to attachMessage as well

app/routing/DeparturesRouter.scala
```diff
@@ -101,4 +101,19 @@ class DeparturesRouter @Inject() (
...
+      runIfBound[V2DepartureId](
+        "departureId",
+        departureId,
+        v2Departures.attachMessage
+      )
+    case _ =>
+      runIfBound[V1DepartureId](
+        "departureId",
+        departureId,
+        v1DepartureMessages.sendMessageDownstream
```
ardwalker, line 115: then you'll need to change this to attachMessage as well

app/controllers/DepartureMessagesController.scala
drnaylor-oc: This is a v1 method, we want to change this as little as possible - this should be left as is.

ardwalker (approved):

PR #301, opened 2022-09-21
Ctcp 1236

derekspalmer:


app/v2/models/responses/MessageResponseWithoutBody.scala
```diff
@@ -0,0 +1,34 @@
...
+
+object MessageResponseWithoutBody {
+  implicit lazy val messageResponseFormatWithoutBody = Json.format[MessageResponseWithoutBody]
+}
+
+case class MessageResponseWithoutBody(
+  id: MessageId,
+  received: OffsetDateTime,
+  messageType: MessageType,
+  body: Option[String] = None
```
liandra-giacomine, line 33: The name MessageResponseWithoutBody doesn't make sense as it does contain the body

derekspalmer: You do have a point :grinning:

Will address.

conf/app.routes
```diff
@@ -19,6 +19,6 @@ GET        /movements/departures/:departureId                       routing.Depa
 POST       /movements/departures/:departureId/messages              routing.DeparturesRouter.attachMessage(departureId: String)
 GET        /movements/departures/:departureId/messages              routing.DeparturesRouter.getMessageIds(departureId: String, receivedSince: Option[OffsetDateTime] ?= None)
 
-GET        /movements/departures/:departureId/messages/:messageId   routing.DeparturesRouter.getMessage(departureId: String, messageId: String)
+GET        /movements/departures/:departureId/message/:messageId    routing.DeparturesRouter.getMessage(departureId: String, messageId: String)
```
drnaylor-oc, line 22: No - this shouldn't be changed.

test/controllers/DepartureMessagesControllerSpec.scala
```diff
@@ -106,7 +106,7 @@ class DepartureMessagesControllerSpec
       |{
       |  "_links": {
       |    "self": {
-      |      "href": "/customs/transits/movements/departures/123/messages/4"
+      |      "href": "/customs/transits/movements/departures/123/message/4"
```
drnaylor-oc, line 109: This should still be `messages`

drnaylor-oc (commented):
It seems the document had a couple of typos that you've copied into here.

app/v2/models/responses/hateoas/HateoasDepartureIdsResponse.scala
```diff
@@ -17,17 +17,30 @@
...
+        response =>
+          Json.obj(
+            "_links" -> Json.obj(
+              "self"     -> departureUri(response._id),
+              "messages" -> s"${departureUri(response._id)}/messages"
+            ),
+            "id"                      -> response._id.value,
+            "movementReferenceNumber" -> response.movementReferenceNumber,
+            "created"                 -> hateoasDateTime.format(response.created),
+            "updated"                 -> hateoasDateTime.format(response.updated),
```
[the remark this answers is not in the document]
derekspalmer: Will change.

app/v2/models/responses/hateoas/HateoasDepartureMessageResponse.scala
```diff
@@ -30,12 +31,10 @@ object HateoasDepartureMessageResponse extends HateoasResponse {
...
       ),
-      "departure" -> Json.obj(
-        "id" -> departureUri(departureId)
-      ),
-      "id"          -> messageUri(departureId, messageId),
-      "received"    -> messageResponse.received.toLocalDateTime,
-      "messageType" -> messageResponse.messageType
+      "id"          -> messageId.value,
+      "departureId" -> departureId.value,
+      "received"    -> CommonFormats.hateoasDateTime.format(messageResponse.received),
```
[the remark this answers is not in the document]
derekspalmer: Will change.

test/controllers/DepartureMessagesControllerSpec.scala
derekspalmer: Will revert

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -785,6 +787,14 @@ class V2DeparturesControllerSpec
     }
   }
 
+  def generateResponse(messageId: MessageId) =
+    MessageResponseWithBody(
+      messageId,
+      arbitrary[OffsetDateTime].sample.value,
+      DepartureDeclaration,
+      Some("<CC015C><test>testxml</test></CC015C>")
+    )
```
RobertLaverick, line 796: You've a few places where you use a MessageResponseWithBody off of a Arbitrary[MessageId] - how about you change them to use an Arbitrary[MessageResponseWithBody] using this inside? Much cleaner.

test/v2/models/responses/hateoas/HateoasDepartureMessageIdsResponseSpec.scala
```diff
@@ -69,12 +84,20 @@ class HateoasDepartureMessageIdsResponseSpec extends AnyFreeSpec with Matchers w
...
     case None => Json.obj("href" -> s"/customs/transits/movements/departures/${departureId.value}/messages")
   }
 
+  private def generateResponse(messageId: MessageId) =
+    MessageResponseWithBody(
+      messageId,
+      arbitrary[OffsetDateTime].sample.value,
+      DepartureDeclaration,
+      Some("<CC015C><test>testxml</test></CC015C>")
+    )
```
RobertLaverick, line 101: DRY principle, see earlier comment - either make this into a trait-method or ideally use that Arbitrary suggestion from before

test/v2/services/DeparturesServiceSpec.scala
```diff
@@ -235,4 +237,12 @@ class DeparturesServiceSpec
...
   }
 
+  private def generateResponse(messageId: MessageId) =
+    MessageResponseWithBody(
+      messageId,
+      arbitrary[OffsetDateTime].sample.value,
+      DepartureDeclaration,
+      Some("<CC015C><test>testxml</test></CC015C>")
+    )
+
```
RobertLaverick, line 247: DRY 2

RobertLaverick (changes requested):

app/v2/models/responses/MessageResponseWithBody.scala
```diff
@@ -0,0 +1,34 @@
...
+object MessageResponseWithBody {
+  implicit lazy val messageResponseFormatWithoutBody = Json.format[MessageResponseWithBody]
+}
+
+case class MessageResponseWithBody(
+  id: MessageId,
+  received: OffsetDateTime,
+  messageType: MessageType,
+  body: Option[String] = None
+)
```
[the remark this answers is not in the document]
derekspalmer: Agree - probably should change the same in transit-movements, as think that's where it's from.

test/v2/controllers/V2DeparturesControllerSpec.scala
derekspalmer: Agree. Will implement.

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -787,13 +785,13 @@ class V2DeparturesControllerSpec
...
-      DepartureDeclaration,
-      Some("<CC015C><test>testxml</test></CC015C>")
-    )
+//  def generateResponse(messageId: MessageId) =
+//    MessageSummary(
+//      messageId,
+//      arbitrary[OffsetDateTime].sample.value,
+//      DepartureDeclaration,
+//      Some("<CC015C><test>testxml</test></CC015C>")
+//    )
```
RobertLaverick, line 794: We don't leave commented code unless there are exceptional reasons.

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -805,7 +803,11 @@ class V2DeparturesControllerSpec
           messageId1 <- arbitrary[MessageId]
           messageId2 <- arbitrary[MessageId]
           messageId3 <- arbitrary[MessageId]
-        } yield Seq(generateResponse(messageId1), generateResponse(messageId2), generateResponse(messageId3))).sample.value
+        } yield Seq(
+          genMessageSummary(Some(messageId1)).arbitrary.sample.value,
+          genMessageSummary(Some(messageId2)).arbitrary.sample.value,
+          genMessageSummary(Some(messageId3)).arbitrary.sample.value
+        )).sample.value
```
RobertLaverick, line 810: I think you missed the point with the use of arbitrary it was to replace the use of the messageId1, messageId2, messageId3 calls not just a relocation of the method you'd written

test/v2/base/CommonGenerators.scala
```diff
@@ -56,4 +60,21 @@ trait CommonGenerators {
...
+      case Some(id) => MessageSummary(id, received, messageType, body)
+      case None =>
+        MessageSummary(
+          genShortUUID.map(MessageId(_)).sample.get,
+          received,
+          messageType,
+          body
+        )
+    }
+  }
```
RobertLaverick, line 78: Every use case of this code uses the Some case for one (so why is it an option?) and only ever directly after generating the Id. When I suggest an arbitrary I meant like the ones above and elsewhere in the repos. i.e. `  implicit lazy val arbitraryMessageWithStateXml: Arbitrary[MessageWithStatus] =
    Arbitrary {
      for {
        messageId   <- arbitrary[MessageId]
        date        <- datesBetween(pastDate, dateNow)
        time        <- timesBetween(pastDate, dateNow)
        xml         <- Gen.const(<blankXml>message</blankXml>)
        messageType <- arbitrary[MessageType]
        status = SubmissionPending
      } yield MessageWithStatus(messageId, LocalDateTime.of(date, time), messageType, xml, status, 1)
    }`

RobertLaverick (changes requested):

test/v2/controllers/V2DeparturesControllerSpec.scala
derekspalmer: Yes. We don't - I'll remove it.

test/v2/controllers/V2DeparturesControllerSpec.scala
derekspalmer: I think I did. Can you provide me with the code you mean?

RobertLaverick: see an example in the comment for CommonGenerators

test/v2/models/responses/hateoas/HateoasDepartureMessageResponseSpec.scala
```diff
@@ -57,12 +55,10 @@ class HateoasDepartureMessageResponseSpec extends AnyFreeSpec with Matchers with
...
       ),
-      "departure" -> Json.obj(
-        "id" -> s"/customs/transits/movements/departures/${departureId.value}"
-      ),
-      "id"          -> s"/customs/transits/movements/departures/${departureId.value}/messages/${messageId.value}",
-      "received"    -> "2022-08-04T11:52:59",
-      "messageType" -> MessageType.DepartureDeclaration.code,
+      "id"          -> messageId.value,
+      "departureId" -> departureId.value,
+      "received"    -> "2022-08-04T11:52:59Z",
```
drnaylor-oc, line 60: I noticed you put the `Z` on the end of the value here in the test, but we've not used that elsewhere. So I did a little digging, because I thought that we specified that we used UTC everywhere... and I actually can't find that, not on our docs, nor on DevHub.

However, in both version 1 of our API, along with this upcoming version 2, we've _not_ been doing that, and I'm keen to keep things as simple as possible when it comes to developers transitioning from v1 to v2. 

I'd not be against raising a spike to decide this point, but right now, the datetime format should remain consistent with what was before -- so please drop the Z for now.

drnaylor-oc: I'm going to write a ticket about this to make this consistent across all calls. The Z shouldn't harm a ISO8601 compliant parser...

app/v2/models/request/MessageType.scala
```diff
@@ -32,10 +32,44 @@ sealed abstract class DepartureMessageType(val code: String) extends MessageType
...
+  case object IE035 extends DepartureMessageType("IE035")
+  case object IE045 extends DepartureMessageType("IE045")
+  case object IE051 extends DepartureMessageType("IE051")
+  case object IE054 extends DepartureMessageType("IE054")
+  case object IE055 extends DepartureMessageType("IE055")
+  case object IE056 extends DepartureMessageType("IE056")
+  case object IE060 extends DepartureMessageType("IE060")
+  case object IE170 extends DepartureMessageType("IE170")
+  case object IE906 extends DepartureMessageType("IE906")
+  case object IE928 extends DepartureMessageType("IE928")
```
RobertLaverick, line 52: We've not used this pattern for the naming of message types elsewhere - the reason being that this completely hides the purpose of the message, and having that context and understanding has value. If you're going to introduce all of these (and I'm not sure this ticket requires it) then use a meaningful name for each.

test/v2/models/responses/hateoas/HateoasDepartureMessageIdsResponseSpec.scala
```diff
@@ -41,25 +47,36 @@ class HateoasDepartureMessageIdsResponseSpec extends AnyFreeSpec with Matchers w
...
-        val actual      = HateoasDepartureMessageIdsResponse(departureId, messageIds, dateTime)
+        val responses = (for {
+          messageId1 <- arbitrary[MessageId]
+          messageId2 <- arbitrary[MessageId]
+          messageId3 <- arbitrary[MessageId]
+        } yield Seq(
+          genMessageSummary(Some(messageId1)).arbitrary.sample.value,
+          genMessageSummary(Some(messageId2)).arbitrary.sample.value,
+          genMessageSummary(Some(messageId3)).arbitrary.sample.value
+        )).sample.value
```
RobertLaverick, line 60: As mentioned elsewhere, you can just use an arbitrary[MessageSummary] call rather than this yielded sequence

RobertLaverick (changes requested):

app/v2/models/request/MessageType.scala
drnaylor-oc: > and I'm not sure this ticket requires it

Unfortunately, it does, at least in the current state. Any of these message types could be sent via the trader or received from ERMIS, which is why I added them now -- if one of these were missing then we get the 500 error.

The other way I was looking at handling this was just to not convert the message type we receive from transit-movements and leave it as a string. It does mean we need to trust ourselves but... maybe that's what we should do?

drnaylor-oc: I've renamed all the constants. The `DepartureMessageType` thing needs to be re-evaluated, but that's not for this PR.

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -229,7 +232,8 @@ class PersistenceConnectorSpec
     "on successful message, return a success" in {
       val eori            = arbitrary[EORINumber].sample.get
       val departureId     = arbitrary[DepartureId].sample.get
-      val messageResponse = okResultGen.sample.get
+      val messageId       = arbitrary[MessageId].sample.get
+      val messageResponse = generateResponseWithoutBody(messageId) //okResultGen.sample.get
```
RobertLaverick, line 236: What's this comment for?

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -113,7 +113,7 @@ class V2DeparturesControllerImpl @Inject() (
         withTemporaryFile {
           (temporaryFile, source) =>
             implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
-            val messageType: MessageType   = MessageType.DepartureDeclaration
+            val messageType: MessageType   = MessageType.DeclarationData
```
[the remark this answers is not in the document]
drnaylor-oc: It's not - it's actually Declaration Data: https://developer.service.hmrc.gov.uk/guides/ctc-traders-phase5-tis/documentation/messagetypes.html#ie015-declaration-data

RobertLaverick (approved):

PR #304, opened 2022-09-26
Optimise labels of CTC Traders API v1 endpoints

fmcgrath:


resources/public/api/conf/1.0/application.raml
```diff
@@ -640,9 +640,9 @@ traits:
               description: When the movement does not exist, has been archived or is not available to the EORI number.
         /{messageId}:
           get:
-            displayName: Get a message relating to a Movement Departure and message ID
+            displayName: Get a cached message related to a movement departure and message ID
```
drnaylor-oc, line 643: Just reading through this... I know this is nothing to do with this PR so much, but movement departure reads weird. I'm not going to block on this, more to flag this up, but surely it should be "departure movement"? A departure is always a movement, so it doesn't need to be modified as "movement departure", but a movement isn't always a departure, so can be.

As I say, it's not something I'm going to block on, this could be a separate PR if we do something with this, but the phraseology is a bit weird.

drnaylor-oc (approved):
Just a musing, not really anything to do with the PR itself...

resources/public/api/conf/1.0/application.raml
RobertLaverick: Personally I'd go with just Departure (for the record) and vice versa with Arrival - it's part of why our routes look like they do. The idea being that "in the context" the Movement is a Departure, or the Movement is an Arrival (and in other offices you could say it is a Transit because they're an office of transit).

But that's my take, not sure if there was any UR done on the terminology in the long-past. 

@crystalmark might have memory of that.

drnaylor-oc (approved):

PR #307, opened 2022-09-28
CTCP-1206: Accept and process XML messages from traders via the API 

sonisn1988:


app/v2/controllers/V2DeparturesController.scala
```diff
@@ -241,9 +243,37 @@ class V2DeparturesControllerImpl @Inject() (
...
+      implicit request =>
+        withTemporaryFile {
+          (temporaryFile, source) =>
+            implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+
+            (for {
+              messageType <- xmlParsingService.validateMessageType(source).asPresentation
+              _           <- validationService.validateXml(messageType, source).asPresentation
+
+              fileSource = FileIO.fromPath(temporaryFile)
```
drnaylor-oc, line 257: This isn't going to work -- `source` can only be used once because it's a request stream.

I was originally not going to merge https://github.com/hmrc/common-transit-convention-traders/pull/299 until this was done because I didn't want to give you too much of a moving target, but it's clear that actually it'd be better if I did. Once I have merged that (it'll be soon), you'll need to update this to use that -- effectively it'll remove the confusing use of multiple sources and you won't need to worry about it.

app/v2/services/AuditingService.scala
```diff
@@ -43,4 +45,17 @@ class AuditingServiceImpl @Inject() (auditingConnector: AuditingConnector) exten
...
+
+      AuditType.values
+        .filterNot(
+          auditType => messageType.equals("DeclarationData")
+        )
+        .find(_.name == messageType) match {
+        case None            => Future.successful(Left(FailedToValidateError.InvalidMessageTypeError(s"Invalid Message Type for auditing : $messageType")))
+        case Some(auditType) => Future.successful(Right(auditType))
+      }
+    }
```
drnaylor-oc, line 60: I don't like this because we're basically using a string comparison to get an AuditType from the same name of the `MessageType`.

Would it not be better to either:

* Make `MessageType` implement `AuditType`, add the audit type string there, and just use that as an audit type, avoid this comparison, or:
* Add an `AuditType` field to `MessageType` so we can just pull the audit type out of the message type?

We can't really get rid of `AuditType`, as I know we'll be auditing failures at some point, and those audit types won't necessarily be specific to a message, but doing this feels error prone.

app/v2/services/MessagesXmlParsingService.scala
```diff
@@ -0,0 +1,64 @@
...
+
+@ImplementedBy(classOf[MessagesXmlParsingServiceImpl])
+trait MessagesXmlParsingService {
+  def validateMessageType(source: Source[ByteString, _]): EitherT[Future, FailedToValidateError, MessageType]
+}
+
+@Singleton
+class MessagesXmlParsingServiceImpl @Inject() (implicit materializer: Materializer) extends MessagesXmlParsingService {
+
+  override def validateMessageType(source: Source[ByteString, _]): EitherT[Future, FailedToValidateError, MessageType] =
```
drnaylor-oc, line 42: You're not validating here, you're extracting the message type, so the name is bad. Equally, you should be using a new left here -- so not `FailedToValidateError` but something else, perhaps `ExtractionError`, and you should create the required cases and typeclass in `ErrorTranslator` to support this.

app/v2/services/MessagesXmlParsingService.scala
```diff
@@ -0,0 +1,64 @@
...
+      source
+        .via(XmlParsing.parser)
+        .mapConcat {
+          case s: StartElement
+              if MessageType.updateDepartureValues
+                .exists(_.rootNode == s.localName) =>
+            Seq(MessageType.updateDepartureValues.find(_.rootNode == s.localName).get)
+          case _ => Seq.empty
+        }
+        .take(1)
```
drnaylor-oc, line 53: If someone passes in XML that does not match any of the message types, (say they submit an arrival message instead of a departure), then it'll run through the entire XML, performing the guard on each element. Similarly, if I was to provide the following:

```xml
<rubbish><a><b></b><CC013C><c></c></CC013C></a></rubbish>
```

this will return CC013C.

This needs to make sure we _only_ read the root node, and tests should reflect this.

app/v2/services/MessagesXmlParsingService.scala
```diff
@@ -0,0 +1,64 @@
...
+        .fold[Either[FailedToValidateError, MessageType]](Left(FailedToValidateError.InvalidMessageTypeError("Message Type")))(
+          (current, next) =>
+            current match {
+              case Left(FailedToValidateError.InvalidMessageTypeError(_)) => Right(next)
+              case _                                                      => Left(FailedToValidateError.UnexpectedError(None))
```
drnaylor-oc, line 58: It's definately not an unexpected error - you know what's happened -- you couldn't extact the message type. Goes hand in hand with the validate comment above.

test/v2/services/MessagesXmlParsingServiceSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+      val source = createStream(validXml)
+
+      val result = service.validateMessageType(source)
+
+      whenReady(result.value) {
+        _ mustBe Right(MessageType.DeclarationAmendment)
+      }
+    }
+
+    "if it doesn't have a valid message type, return ParseError.MessageTypeNotFound" in {
```
drnaylor-oc, line 65: The name doesn't match the error.

Additionally, add a test for what I said above.

test/v2/services/MessagesXmlParsingServiceSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+class MessagesXmlParsingServiceSpec
+    extends AnyFreeSpec
+    with TestActorSystem
+    with Matchers
+    with StreamTestHelpers
+    with ScalaFutures
+    with ScalaCheckPropertyChecks {
+
+  val validXml: NodeSeq =
+    <CC013C>
```
drnaylor-oc, line 39: I'm trying to standardise this across everything, as the actual root node that comes from a trader will look like:

```xml
<ncts:CC013C PhaseID="NCTS5.0" xmlns:ncts="http://ncts.dgtaxud.ec">...</ncts:CC013C>
```

so please use that in your tests, rather than just the root name (it shouldn't affect the tests, but it's better to have this so that if something changes, our tests will fail and we'll know about it).

app/v2/services/AuditingService.scala
```diff
@@ -43,4 +45,17 @@ class AuditingServiceImpl @Inject() (auditingConnector: AuditingConnector) exten
       case NonFatal(e) =>
         logger.warn("Unable to audit payload due to an exception", e)
     }
+
+  def getAuditType(messageType: String): EitherT[Future, FailedToValidateError, AuditType] =
+    EitherT {
+
+      AuditType.values
+        .filterNot(
+          auditType => messageType.equals("DeclarationData")
```
drnaylor-oc, line 54: Why?

drnaylor-oc (commented):
Some musings - the structure is there but there are a few things to think about.

app/v2/services/AuditingService.scala
```diff
@@ -43,4 +45,17 @@ class AuditingServiceImpl @Inject() (auditingConnector: AuditingConnector) exten
...
+        )
+        .find(_.name == messageType) match {
+        case None            => Future.successful(Left(FailedToValidateError.InvalidMessageTypeError(s"Invalid Message Type for auditing : $messageType")))
```
drnaylor-oc, line 57: This again is not a validation error, so we should not be using `FailedToValidateError`s here.

app/v2/controllers/V2DeparturesController.scala
sonisn1988: done the changes

app/v2/services/AuditingService.scala
sonisn1988: done the changes

app/v2/services/AuditingService.scala
sonisn1988: removed

app/v2/services/MessagesXmlParsingService.scala
sonisn1988: changed method name and added Error Translator

app/v2/services/AuditingService.scala
sonisn1988: removed complete method from the Audit service not required

app/v2/services/MessagesXmlParsingService.scala
sonisn1988: done

test/v2/services/MessagesXmlParsingServiceSpec.scala
sonisn1988: done

app/v2/services/MessagesXmlParsingService.scala
sonisn1988: done

test/v2/services/MessagesXmlParsingServiceSpec.scala
sonisn1988: done

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -81,8 +81,8 @@ class PersistenceConnectorSpec
     with ScalaCheckDrivenPropertyChecks
     with CommonGenerators {
 
-  lazy val appConfig: AppConfig = app.injector.instanceOf[AppConfig]
-
+  lazy val appConfig: AppConfig                           = app.injector.instanceOf[AppConfig]
+  lazy val messageTypeGen                                 = Gen.oneOf(Seq(MessageType.DeclarationAmendment))
```
RobertLaverick, line 85: If this will over produce a DeclarationAmendment, why is it a generator?

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -659,6 +659,135 @@ class PersistenceConnectorSpec
       Some("<CC015C><test>testxml</test></CC015C>")
     )
 
+  "POST /traders/movements/:movementId/messages" - {
+    lazy val okResultGen =
+      for {
+        messageId <- arbitrary[MessageId]
+      } yield messageId
```
RobertLaverick, line 666: That name is misleading, if this generates a messageId it should at least be called that for when it's seen elsewhere.

Also, I'm pretty sure you could just do `val messageIdGen = arbitrary[MessageId]`

test/v2/connectors/V2BaseConnectorSpec.scala
```diff
@@ -132,6 +132,13 @@ class V2BaseConnectorSpec
       urlPath.toString() mustBe s"/transit-movements-converter/messages/${messageType.code}"
   }
 
+  "the post departure movements URL on localhost for a given departureId should be as expected" in {
+    val departureId = arbitrary[DepartureId].sample.value
+    val urlPath     = Harness.movementsPostDeparture(departureId)
+
+    urlPath.toString() mustBe s"/transit-movements/traders/movements/${departureId.value}/messages/"
+  }
+
```
RobertLaverick, line 141: strictly that endpoint is for any movement, not just departure movements. (At least it will be in the future). Not sure if we *need* to rename it, but just an observation.

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -159,6 +179,14 @@ class V2DeparturesControllerSpec
   ): Request[A] =
     FakeRequest(method = method, uri = uri, headers = headers, body = body)
 
+  def fakeAttachDepartures[A](
+    method: String,
+    headers: FakeHeaders,
+    uri: String = routing.routes.DeparturesRouter.attachMessage("123").url,
```
RobertLaverick, line 185: if this is always going to be for attachDepartures, you don't need the url to be a parameter

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -1083,6 +1115,99 @@ class V2DeparturesControllerSpec
     }
   }
 
+  // Version 2
```
RobertLaverick, line 1118: what's this comment for? this entire spec is V2

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -1083,6 +1115,99 @@ class V2DeparturesControllerSpec
...
+  "for a departure update with accept header set to application/vnd.hmrc.2.0+json (version two)" - {
```
RobertLaverick, line 1119: Can you please make this resemble the other test descriptions in the spec?

test/v2/services/DeparturesServiceSpec.scala
```diff
@@ -229,4 +229,33 @@ class DeparturesServiceSpec
     }
 
   }
+
+  "When Updating departure with departureId and messageType" - {
```
RobertLaverick, line 233: You can drop the When here, it breaks with the style of the spec

test/v2/services/MessagesXmlParsingServiceSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+    </ncts:CC013C>
+
+  val invalidMessageType: NodeSeq =
+    <CC015C>
+      <CC013C>
+        <identificationNumber>GB1234</identificationNumber>
+      </CC013C>
+    </CC015C>
+
+  "When handed an XML stream" - {
```
RobertLaverick, line 52: We typically use a test name scheme of the method name being described, this is very vague. Perhaps something like "extractMessageType" and then "if stream is valid...."

test/v2/services/XmlParsersSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
+import v2.base.StreamTestHelpers
+import v2.base.TestActorSystem
+import v2.models.errors.ExtractionError
+import v2.models.request.MessageType
+
+import scala.xml.NodeSeq
+
+class XmlParsersSpec extends AnyFreeSpec with TestActorSystem with Matchers with StreamTestHelpers with ScalaFutures with ScalaCheckPropertyChecks {
+  "MessageType parser" - {
```
RobertLaverick, line 32: If you're calling it a MessageTypeExtractor, use that term please so people can search for that term through the tests

RobertLaverick (changes requested):

app/v2/models/request/MessageType.scala
```diff
@@ -25,50 +25,59 @@ import play.api.libs.json.Writes
 sealed trait MessageType {
   def code: String
   def movementType: String
+  def rootNode: String
+  def name: String
 }
```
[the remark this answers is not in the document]
sonisn1988: this is already done. Please check the latest changes with this branch

app/v2/models/request/MessageType.scala
```diff
@@ -25,50 +25,59 @@ import play.api.libs.json.Writes
...
     NoReleaseForTransit,
+    PositiveAcknowledge,
     GuaranteeNotValid,
     RejectionFromOfficeOfDeparture,
     ControlDecisionNotification,
-    PresentationNotificationForThePreLodgedDecision,
     FunctionalNack,
     PositiveAcknowledge
   )
+  val values = updateDepartureValues ++ departureValues
```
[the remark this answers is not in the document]
sonisn1988: we need to validate only 4 specific message type with as per update departure so collected in one collection
https://confluence.tools.tax.service.gov.uk/display/CIP/CTC+Traders+API+Phase+5+-+CIP+Assessment
based on above guidelines for sent by departure message

app/v2/connectors/PersistenceConnector.scala
```diff
@@ -75,6 +76,10 @@ trait PersistenceConnector {
     ec: ExecutionContext
   ): Future[Seq[DepartureResponse]]
 
+  def post(departureId: DepartureId, messageTypeCode: String, source: Source[ByteString, _])(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): Future[MessageId]
```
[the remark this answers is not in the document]
sonisn1988: done

app/v2/services/DeparturesService.scala
```diff
@@ -67,6 +67,11 @@ trait DeparturesService {
     ec: ExecutionContext
   ): EitherT[Future, PersistenceError, Seq[DepartureResponse]]
 
+  def updateDeparture(departureId: DepartureId, messageTypeCode: String, source: Source[ByteString, _])(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): EitherT[Future, PersistenceError, MessageId]
+
```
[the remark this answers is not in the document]
sonisn1988: done

it/v2/connectors/PersistenceConnectorSpec.scala
sonisn1988: done the changes

it/v2/connectors/PersistenceConnectorSpec.scala
sonisn1988: done the changes

test/v2/connectors/V2BaseConnectorSpec.scala
sonisn1988: Keeping it is as of now. Will change later if required

test/v2/controllers/V2DeparturesControllerSpec.scala
sonisn1988: removed

test/v2/controllers/V2DeparturesControllerSpec.scala
sonisn1988: done

test/v2/services/DeparturesServiceSpec.scala
sonisn1988: done

test/v2/services/MessagesXmlParsingServiceSpec.scala
sonisn1988: done

test/v2/services/XmlParsersSpec.scala
sonisn1988: done

test/v2/controllers/V2DeparturesControllerSpec.scala
sonisn1988: That is required as FakeRequst without url does not comply

app/v2/models/request/MessageType.scala
```diff
@@ -21,54 +21,66 @@ import play.api.libs.json.JsString
...
+    AmendmentAcceptance,
+    InvalidationDecision,
     DeclarationData,
     Discrepancies,
     MRNAllocated,
     ReleaseForTransit,
     RecoveryNotification,
     WriteOffNotification,
     NoReleaseForTransit,
+    PositiveAcknowledge,
```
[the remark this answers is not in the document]
sonisn1988: done

app/v2/services/XmlParsers.scala
```diff
@@ -0,0 +1,45 @@
...
+        true
+      case _ => false
+    }
+    .take(1)
+    .map {
+      _ match {
+        case s: StartElement =>
+          if (MessageType.updateDepartureValues.exists(_.rootNode == s.localName)) {
+            Right(MessageType.values.find(_.rootNode == s.localName).get)
+          } else Left(ExtractionError.MessageTypeNotFound(s.localName))
```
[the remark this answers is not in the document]
sonisn1988: done as suggested

app/v2/services/XmlParsers.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  val messageTypeExtractor: Flow[ParseEvent, Either[ExtractionError, MessageType], NotUsed] = Flow[ParseEvent]
+    .filter {
+      case _: StartElement =>
+        true
+      case _ => false
+    }
+    .take(1)
+    .map {
+      _ match {
```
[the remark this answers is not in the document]
sonisn1988: done. Thanks Dan

test/v2/services/MessagesXmlParsingServiceSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+    "if it doesn't have a valid message type, return ExtractionError.MessageTypeNotFound" in {
+      val xmlParsingService = new MessagesXmlParsingServiceImpl()
+      val payload           = createStream(invalidMessageType)
+      val response =
+        xmlParsingService.extractMessageType(payload)
+
+      whenReady(response.value) {
+        _ mustBe Left(ExtractionError.MessageTypeNotFound("CC015C"))
+      }
+    }
```
RobertLaverick, line 73: You should include a case where there are *no* CC tags for completeness

RobertLaverick (commented):
One simple new test and then I'm happy to approve this

test/v2/services/MessagesXmlParsingServiceSpec.scala
sonisn1988: added

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -81,22 +81,22 @@ class PersistenceConnectorSpec
...
-
+  lazy val appConfig: AppConfig                           = app.injector.instanceOf[AppConfig]
+  lazy val messageType                                    = MessageType.DeclarationAmendment
   lazy val persistenceConnector: PersistenceConnectorImpl = new PersistenceConnectorImpl(httpClientV2, appConfig, new TestMetrics())
   implicit lazy val ec: ExecutionContext                  = app.materializer.executionContext
 
   "POST /traders/:eori/movements/departures" - {
 
-    lazy val okResultGen =
+    lazy val okResult =
```
drnaylor-oc, line 91: More a comment, but I'm unsure as to why you renamed this. The reason it was `okResultGen` is because it returns a `Gen[DeclarationResponse]`, and that `okResult` is used in the lambda in line 99, which then shadows this declaration.

It's not a huge problem as this variable shouldn't be used within the test itself, but in general I'd recommend not shadowing variables, it just gets confusing if you do.

sonisn1988: reverted back to original

drnaylor-oc:
I just tested this using our API regression test suite with https://github.com/hmrc/transit-movements-router/pull/14 and the rest of the services and the tests failed. One of the requirements for this task is:

> Set the message type as IEXXX as the header X-Message-Type when sending messages to transit-movements and transit-movements-router

You've just missed this on sending to the router (and to transit-movements for posting a new departure, but that doesn't cause an error).

Once you've done that, ensure that you run this with [the API regression test suite](https://github.com/hmrc/ctc-traders-api-tests) and [this router branch](https://github.com/hmrc/transit-movements-router/pull/14) to make sure everything works as expected.

app/v2/connectors/RouterConnector.scala
```diff
@@ -64,7 +65,7 @@ class RouterConnectorImpl @Inject() (val metrics: Metrics, appConfig: AppConfig,
 
         httpClientV2
           .post(url"$url")
-          .addHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.XML)
+          .addHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.XML, Constants.XMessageTypeHeader -> messageType.code)
```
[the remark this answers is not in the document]
drnaylor-oc: My PR was meant as a stopgap to allow us to merge https://github.com/hmrc/transit-movements-router/pull/14 if we wanted to get that over the line. The general intention was for this PR to be the trigger, however, and that PR will be closed without merging if this is pulled before.

The implementation is the same anyway - but yes, I would like the tests to match. We know that this header is required, so our tests should be ensuring that.

RobertLaverick (approved):

PR #311, opened 2022-10-04
added 404 error check for departure does not exist

sonisn1988:


app/v2/models/errors/PersistenceError.scala
```diff
@@ -25,6 +25,7 @@ sealed trait PersistenceError
 object PersistenceError {
   case class MessageNotFound(movementId: DepartureId, messageId: MessageId) extends PersistenceError
   case class DepartureNotFound(departureId: DepartureId)                    extends PersistenceError
+  case class DepartureDoesNotExist(departureId: DepartureId)                extends PersistenceError
```
drnaylor-oc, line 28: No need to add this, use `DepartureNotFound`

drnaylor-oc (commented):
Use what's already there.

app/v2/models/errors/PersistenceError.scala
sonisn1988: That is mapped with 404 in Error Translator

sonisn1988: so created new one and mapped with 400

drnaylor-oc: It's `ErrorTranslator` we need to make a modification to then. We already do something there for an override:

https://github.com/hmrc/common-transit-convention-traders/blob/1ee9f7cb1235f651e4b95fa99d909fb8da3d47b5/app/v2/controllers/ErrorTranslator.scala#L47-L64

and then use it explicitly

https://github.com/hmrc/common-transit-convention-traders/blob/1ee9f7cb1235f651e4b95fa99d909fb8da3d47b5/app/v2/controllers/V2DeparturesController.scala#L135

and so we should do something similar here if we want to override, because it's the controller that cares, not the rest of the stack.

drnaylor-oc:
After discussion, we realised 404 was actually the right answer -- sorry! However, you said you need to make a change anyway, so we'll keep this open for that.

app/v2/models/errors/PersistenceError.scala
sonisn1988: As discussed in clarification not required

drnaylor-oc (approved):

PR #313, opened 2022-10-05
Bug fixes for 415 and invalid request body

sonisn1988:


app/v2/controllers/ErrorTranslator.scala
```diff
@@ -92,9 +86,9 @@ trait ErrorTranslator {
 
   implicit val extractionError = new Converter[ExtractionError] {
 
-    def convert(extractionError: ExtractionError): PresentationError = extractionError match {
-      case err: UnexpectedError             => PresentationError.internalServiceError(cause = err.thr)
-      case MessageTypeNotFound(messageType) => PresentationError.badRequestError(s"$messageType is not a valid message type")
+    override def convert(extractionError: ExtractionError): PresentationError = extractionError match {
+      case ExtractionError.InvalidInputType()               => PresentationError.badRequestError("Invalid Input type error")
```
drnaylor-oc, line 90: Your proposed error message doesn't read very well. I would turn this into a `MalformedInput` object with the message "Input was malformed" - at that point it's clear the XML is malformed and that's what the trader needs to look at.

sonisn1988: Thanks Dan. Done the changes as suggested

drnaylor-oc (approved):

PR #314, opened 2022-10-10
CTCP-1322: Add push-notification association call, rename DepartureId -> MovementId

drnaylor-oc:


derekspalmer (approved):

test/v2/controllers/V2DeparturesControllerSpec.scala
```diff
@@ -1165,6 +1174,7 @@ class V2DeparturesControllerSpec
           mockDeparturesPersistenceService,
           mockRouterService,
           mockAuditService,
+          mockPushNotificationService,
           FakeMessageSizeActionProvider,
```
liandra-giacomine, line 1178: Need to verify what gets returned when a PushNotificationError error is thrown

app/v2/connectors/V2BaseConnector.scala
```diff
@@ -105,17 +114,21 @@ trait V2BaseConnector extends HttpErrorFunctions {
...
           response =>
             response.status match {
-              case ACCEPTED => Future.successful(())
-              case _        => response.error
+              case `expected` => Future.successful(())
+              case _          => response.error
             }
         }
 
+    def executeAccepted(implicit ec: ExecutionContext): Future[Unit] = expectStatus(ACCEPTED)
```
liandra-giacomine, line 128: Rather than creating a function for each status code, would it make sense to make expectStatus public and call that instead?

app/v2/services/MessagesXmlParsingService.scala
```diff
@@ -74,7 +74,7 @@ class MessagesXmlParsingServiceImpl @Inject() (implicit materializer: Materializ
         .toMat(messageTypeSink)(Keep.right)
         .run()
         .recover {
-          case NonFatal(e) => Left(ExtractionError.MalformedInput())
+          case NonFatal(_) => Left(ExtractionError.MalformedInput)
```
liandra-giacomine, line 77: In the tests, can we verify this ExtractionError gets returned when a malformed input is supplied?

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -87,7 +87,7 @@ trait ErrorTranslator {
   implicit val extractionError = new Converter[ExtractionError] {
```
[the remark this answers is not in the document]
drnaylor-oc: We don't, as we don't actually use the result in returning the status to the use (we fire and forget).

Good question though -- should we be failing the call if an association fails? I suppose it can fail for not having a box at all (rare, but possible, I guess), or just an overall failure.

test/v2/controllers/V2DeparturesControllerSpec.scala
drnaylor-oc: Nothing does, it's currently fire and forget.

app/v2/services/MessagesXmlParsingService.scala
drnaylor-oc: Do we not have that? This was just an adjustment to a previous PR to make a case class a case object, as it took no inputs, so there should be tests for that already.

liandra-giacomine: Don't think there are though, should've been done as part of the previous PR but thought I'd mention here since I couldn't see any

drnaylor-oc: I shall take a look!

sonisn1988:
FYI...I have not included this as part of arrival notification development. Please include this new service call while creating arrival notification

drnaylor-oc:
Ooh, yes, thanks for pointing that out!

test/v2/services/MessagesXmlParsingServiceSpec.scala
```diff
@@ -89,5 +90,17 @@ class MessagesXmlParsingServiceSpec
         _ mustBe Left(ExtractionError.MessageTypeNotFound("HolderOfTheTransitProcedure"))
       }
     }
+
+    "if the input is malformed, return EctractionError.MalformedInput" in {
```
liandra-giacomine, line 94: EctractionError -> ExtractionError :')

liandra-giacomine (approved):

test/v2/services/MessagesXmlParsingServiceSpec.scala
drnaylor-oc: I'll fix that... got the arrivals stuff to shoe in too.

drnaylor-oc:
Not sure why that failed -- worked locally, so re-running to see what happens.

drnaylor-oc:
That passed. Weird.

liandra-giacomine (dismissed):

app/v2/models/MovementType.scala
```diff
@@ -0,0 +1,46 @@
...
+import play.api.libs.json.Writes
+
+sealed abstract class MovementType(val movementType: String)
+
+object MovementType {
+  case object Departure extends MovementType("departure")
+  case object Arrival   extends MovementType("arrival")
+
+  implicit val movementTypeReads: Reads[MovementType] = Reads {
+    case x: JsString => MovementType.findByName(x.value).map(JsSuccess(_)).getOrElse(JsError())
```
liandra-giacomine, line 32: Would writing it something like this read better?

`case JsString(movementType) if MovementType.findByName(movementType) => JsSuccess(movementType)
case _ => JsError()`

app/v2/models/MovementType.scala
```diff
@@ -0,0 +1,46 @@
...
+import play.api.libs.json.JsError
+import play.api.libs.json.JsString
+import play.api.libs.json.JsSuccess
+import play.api.libs.json.Reads
+import play.api.libs.json.Writes
+
+sealed abstract class MovementType(val movementType: String)
+
+object MovementType {
+  case object Departure extends MovementType("departure")
```
sonisn1988, line 28: This should be arrivals and departures ? or arrival and departure. I am thinking we need to append s here

app/v2/models/MovementType.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import play.api.libs.json.JsError
+import play.api.libs.json.JsString
+import play.api.libs.json.JsSuccess
+import play.api.libs.json.Reads
+import play.api.libs.json.Writes
+
+sealed abstract class MovementType(val movementType: String)
+
+object MovementType {
```
sonisn1988, line 27: can we not use movementType string available in MessageType trait? not sure if we need to create this new class if we have something available with messageType

app/v2/models/MovementType.scala
liandra-giacomine: Oh yes, my bad.

app/v2/models/MovementType.scala
drnaylor-oc: Quite the opposite really, the `MessageType` should really contain a `MovementType` as the movement type has become used in more than one place -- so I'll be updating it that way.

There is the problem where _that_ is a plural, but that's down to REST conventions and is correct in that instance, so the movement type will have that pluralised version as an additional field (but you'll still get the singular version).

liandra-giacomine (dismissed):

app/config/CTCServicesConfig.scala
```diff
@@ -0,0 +1,31 @@
...
+
+/*
+ * This class exists simply to expose the config function, which allows us to
+ * avoid a magic string already present in the ServicesConfig object.
+ */
+class CTCServicesConfig @Inject() (config: Configuration) extends ServicesConfig(config) {
+
+  override def config(serviceName: String): Configuration = super.config(serviceName)
+
+}
```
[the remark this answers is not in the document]
drnaylor-oc: [I added a comment to the class to explain it](https://github.com/hmrc/common-transit-convention-traders/pull/314/files/47fc965442421acf5ecde83b660bc0bd4adec5af#diff-3ebf2055d88ed788e8aa3a1b7f1f8cea47035585bba0fd3e75a0c82de3ff5a22R23-R26) -- it's to get access to the protected method `config` so that we don't have to replicate getting the config for a specific key.

PR #315, opened 2022-10-12
refactored departureId with MovementId and XML arrival notification

sonisn1988:


app/controllers/ArrivalMovementController.scala
```diff
@@ -42,9 +41,15 @@ import utils.CallOps._
...
+import java.time.OffsetDateTime
+import javax.inject.Inject
 import scala.concurrent.ExecutionContext
 import scala.xml.NodeSeq
-import controllers.actions.AnalyseMessageActionProvider
+
+@ImplementedBy(classOf[ArrivalMovementController])
+trait V1ArrivalMovementController {
+  def createArrivalNotification(): Action[NodeSeq]
+}
```
liandra-giacomine, line 53: make this a @Singleton

liandra-giacomine (commented):
Looks good, just one minor thing from me

app/controllers/ArrivalMovementController.scala
sonisn1988: Done the changes as suggested. Thanks Liandra

drnaylor-oc:
I've just updated the regression test pack and it now works -- so can you please run the regression tests against it and make sure it works before we merge it?

I know the arrivals haven't been written yet, but the change is invasive enough I just want to have confidence that everything else still works! (It looks like it will, just want to verify that)

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -790,7 +791,136 @@ class PersistenceConnectorSpec
...
+        whenReady(future) {
+          result =>
+            result.left.get mustBe a[UpstreamErrorResponse]
+            val response = result.left.get.asInstanceOf[UpstreamErrorResponse]
+            response.statusCode mustBe BAD_REQUEST
+            Json.parse(response.message).validate[StandardError] mustBe JsSuccess(StandardError("Bad request", ErrorCode.BadRequest))
+        }
+    }
+
+    "On an incorrect Json fragment, must return a JsResult.Exception" in forAll(arbitrary[EORINumber]) {
```
drnaylor-oc, line 892: ...but it's returning a JsonParseException (though I remember writing this for departures, did I get it wrong there?)

Could we update the text to say:

"On an incorrect Json fragment from transit-moments, must return a JsonParseException"?

Just makes it more explicit!

resources/public/api/conf/2.0/application.raml
```diff
@@ -372,4 +372,40 @@ traits:
...
+                {
+                "_links":{
+                "self":{
+                "href": "/customs/transits/movements/arrivals/62fb8f33d233578e"
+                },
+                "messages":{
+                "href": "/customs/transits/movements/arrivals/62fb8f33d233578e/messages"
+                }
+                }
+                }
```
drnaylor-oc, line 411: Example is fine, but can we make that it's indented properly please?

resources/public/api/conf/2.0/application.raml
```diff
@@ -372,4 +372,40 @@ traits:
...
+    post:
+      displayName: Send an arrival notification message
+      description: |
+        Notify the Customs Office of Destination that a transit movement has arrived at its final destination.
+
+        This request is also called an ‘Arrival Notification’ E_ARR_NOT (IE007) message.
+
+        A successful response will contain a URI and an arrival ID for the notification, which you can use to retrieve the message status.
+
+        It will also contain an embedded request ID, box ID and box name, which you can use to match this movement to any messages received from the Push Pull Notifications API.
```
drnaylor-oc, line 386: No it doesn't, not this time -- traders can (and will) get this themselves. Remove this line.

resources/public/api/conf/2.0/application.raml
```diff
@@ -372,4 +372,40 @@ traits:
...
+                          code: NOT_FOUND
+  /arrivals:
+    post:
+      displayName: Send an arrival notification message
+      description: |
+        Notify the Customs Office of Destination that a transit movement has arrived at its final destination.
+
+        This request is also called an ‘Arrival Notification’ E_ARR_NOT (IE007) message.
+
+        A successful response will contain a URI and an arrival ID for the notification, which you can use to retrieve the message status.
```
drnaylor-oc, line 384: Delete "and an arrival ID", we're not returning that (right this second)

drnaylor-oc (commented):
Generally okay - just nitpicky on a name, and the API definition needs tweaking.

Once this is okay, co-ordinate with @RobertLaverick about the order to merge this in please.

it/v2/connectors/PersistenceConnectorSpec.scala
sonisn1988: done the changes

resources/public/api/conf/2.0/application.raml
sonisn1988: done

resources/public/api/conf/2.0/application.raml
sonisn1988: removed

resources/public/api/conf/2.0/application.raml
sonisn1988: removed

app/v2/services/RouterService.scala
```diff
@@ -56,6 +58,8 @@ class RouterServiceImpl @Inject() (routerConnector: RouterConnector) extends Rou
           _ => Right(())
         )
         .recover {
+          case UpstreamErrorResponse(_, BAD_REQUEST, _, _) =>
+            Left(RouterError.UnrecognisedOffice("Unrecognised Office either should start with GB or XI"))
```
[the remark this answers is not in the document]
sonisn1988: done the changes as suggested

app/v2/models/errors/RouterError.scala
```diff
@@ -20,4 +20,5 @@ sealed trait RouterError
 
 object RouterError {
   case class UnexpectedError(thr: Option[Throwable] = None) extends RouterError
+  case class UnrecognisedOffice(message: String)            extends RouterError
```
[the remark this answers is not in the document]
sonisn1988: done the changes

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -74,6 +75,7 @@ trait ErrorTranslator {
 
     override def convert(routerError: RouterError): PresentationError = routerError match {
       case err: RouterError.UnexpectedError => PresentationError.internalServiceError(cause = err.thr)
+      case UnrecognisedOffice(message)      => PresentationError.badRequestError(message)
```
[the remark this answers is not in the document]
sonisn1988: done the changes. will speak to Fergal if we need to change something else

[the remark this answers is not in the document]
sonisn1988: I discussed this with Fergal and he has suggested more generic way to do this. see below:

The customs office specified for `CustomsOfficeOfDestinationActual` or `CustomsOfficeOfDeparture` must be a customs office located in the United Kingdom.

His intention was to not use brackets and slash.

Please suggest if you are happy with above message? so that I can change it as per above suggestion given by Fergal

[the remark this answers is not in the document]
drnaylor-oc: Well, not really. I think the context of what I was saying was a bit lost here and .

The problem is that either `CustomsOfficeOfDeparture \ referenceNumber` or `CustomsOfficeOfDestinationActual \ referenceNumber` (the XML path to the element) must be in the UK, but it's not quite an either or, it's totally dependent on whether this is a departure (`CustomsOfficeOfDeparture`) or an arrival (`CustomsOfficeOfDestinationActual`). 

I'd written it the way I did as I was (badly!) trying to infer it's one of the other depending on this. We'll possibly need to have two error converters with two different messages for this reason.

Need to think on this tomorrow.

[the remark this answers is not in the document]
sonisn1988: I think we can display same messages whatever we receive from transit-movements-router in case of bad request to common-transit-convention-traders as well.
in our case with pending pull request of router I can see : **Did not recognise office: ${office.value}**
Better we can collect above same message and display with bad request response ? It's having users entered bad value so user should be able to identify what went wrong and then user can correct that?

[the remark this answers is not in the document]
drnaylor-oc: The problem with the router's message is that it doesn't specify which office went wrong, and this is going to be important particularly for arrivals, where they might have specified the office of destination and the office of destination actual as the same. The only one we care about for arrivals is the office of destination actual. Really, we need to specify the right one at the right time.

It's generally _not_ a good idea to pass through error messages from internal components (router) to API consumers because we do not want to accidently expose some internal messages. `common-transit-convention-traders` job is to turn internal error messages into ones that are appropriate for the user. If the router PR needs changes to support what you want to do, such as adding fields to the bad request response at the router level, feel free to ask for that! However, I don't want to pass through error messages unchecked, like you're proposing here.

For what it's worth, I've opened CTCP-1467 to put this check in a better place, but agreeing the wording now still makes sense.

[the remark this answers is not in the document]
sonisn1988: I think we can make decision based on the movementType (arrivals or departures) as we are passing movementType at the router level? 
anyway, do we need to put this on hold until decision is made using CTCP-1467?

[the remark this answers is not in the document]
drnaylor-oc: It depends on how quickly 1467 is going to come in and how we do it. I would have liked to discuss it in clarification today, alas, I have another meeting then.

Assume that 1467 isn't being done for now. As I said, the wording is important. And yes, we can use the movement type, but the message needs to be generated here.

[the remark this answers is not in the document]
sonisn1988: As discussed with Rob as of now I have made it generic message as suggested by Fergal.
I have created one ticket to track this issue CTCP-1505. That is dependent of CTCP-1467. Once CTCP-1467 is done. we will further decide what needs to be done as part of  CTCP-1505

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -75,7 +75,10 @@ trait ErrorTranslator {
 
     override def convert(routerError: RouterError): PresentationError = routerError match {
       case err: RouterError.UnexpectedError => PresentationError.internalServiceError(cause = err.thr)
-      case UnrecognisedOffice(message)      => PresentationError.badRequestError(message)
+      case UnrecognisedOffice =>
+        PresentationError.badRequestError(
+          "The customs office specified at ([\"CustomsOfficeOfDestinationActual\" or \"CustomsOfficeOfDeparture\"]\"referenceNumber\" must be a office in the UK"
```
liandra-giacomine: "an office" rather than "a office"

liandra-giacomine (approved):

PR #319, opened 2022-10-20
Trader can send an arrival notification (Json) 

liandra-giacomine:


sonisn1988 (dismissed):
LGTM

app/v2/controllers/V2ArrivalsController.scala
```diff
@@ -79,14 +81,29 @@ class V2ArrivalsControllerImpl @Inject() (
...
+  def createArrivalNotificationJSON(): Action[Source[ByteString, _]] =
+    (authActionNewEnrolmentOnly andThen messageSizeAction()).stream {
+      implicit request =>
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+
+        (for {
+          _ <- validationService.validateJson(MessageType.ArrivalNotification, request.body).asPresentation
+          _ = auditService.audit(AuditType.ArrivalNotification, request.body, MimeTypes.JSON)
+          xmlSouce                  <- conversionService.jsonToXml(MessageType.ArrivalNotification, request.body).asPresentation
+          arrivalNotificationResult <- persistAndSend(xmlSouce)
```
drnaylor-oc, line 100: I'm not sure this'll work as you think, because the `xmlSource` will be a one-use source, a bit like the request stream - we'll need this to use `withReusableSource` like Rob does in his PR.

test/v2/controllers/V2ArrivalsControllerSpec.scala
```diff
@@ -210,109 +218,265 @@ class V2ArrivalsControllerSpec
...
+            .validateJson(eqTo(MessageType.ArrivalNotification), any[Source[ByteString, _]]())(any[HeaderCarrier], any[ExecutionContext])
+        ).thenAnswer {
+          invocation =>
+            jsonValidationMockAnswer(invocation)
+        }
+        when(mockAuditService.audit(any(), any(), eqTo(MimeTypes.JSON))(any(), any())).thenReturn(Future.successful(()))
+
+        val jsonToXmlConversion = (invocation: InvocationOnMock) =>
+          EitherT.rightT(
+            invocation.getArgument[Source[ByteString, _]](1)
```
drnaylor-oc, line 366: Rather than just returning the same thing, we should make this use a `singleUseSource` to emulate the behaviour of what is returned -- I'd imagine this would result in a test failure then for the reason I gave above.

Given this is unit testing, I don't think it matters what is returned, as we're controlling everything!

app/v2/controllers/V2ArrivalsController.scala
drnaylor-oc: https://github.com/hmrc/common-transit-convention-traders/pull/317/files#diff-4530c97418bd3c5db8f389bc12c415848247c437e9393aa9b85dc562227e656dR252-R262 is one way to do it.

liandra-giacomine: For now we only need it once right? I thought it would be something to refactor when PPNS is added?

liandra-giacomine: Oh of course! My bad...

app/v2/controllers/V2ArrivalsController.scala
```diff
@@ -79,21 +84,50 @@ class V2ArrivalsControllerImpl @Inject() (
...
+              withReusableSource(src) {
+                xmlSource =>
+                  for {
+                    _ <- validationService
+                      .validateXml(MessageType.ArrivalNotification, xmlSource)
+                      .asPresentation(jsonToXmlValidationErrorConverter, materializerExecutionContext)
+                    arrivalResult <- arrivalsService.createArrival(request.eoriNumber, xmlSource).asPresentation
+                    _ <- routerService
+                      .send(MessageType.ArrivalNotification, request.eoriNumber, arrivalResult.arrivalId, arrivalResult.messageId, xmlSource)
+                      .asPresentation
```
drnaylor-oc, line 114: I think we should keep `persistAndSend` as it was to avoid the duplication, because the source you'll pass will still work.

drnaylor-oc (approved):

PR #322, opened 2022-10-21
CTCP-1333: Added PPNS configuration for V2

liandra-giacomine:


prashanthmamidi (approved):

drnaylor-oc (approved):
I think that this is all that is needed for getting PPNS set up. We should check on QA once this has been built to see if a) it prompts us for it, and b) if we can get the box (even if via a manual check).

PR #325, opened 2022-10-24
Optimise CTC Traders API v2 response descriptions

fmcgrath:


resources/public/api/conf/2.0/application.raml
```diff
@@ -324,7 +324,7 @@ traits:
                   type: types.errorResponse
                   examples:
                     badRequest:
-                      description: When sent by departure message type not available within xml or xml schema validation failed
+                      description: The departure message type is not available within XML or the message failed schema validation.
```
derekspalmer, line 327: ... within the XML ...

derekspalmer (approved):

resources/public/api/conf/2.0/application.raml
drnaylor-oc: I wouldn't even do that. "The departure message type is not available within XML" is itself part of the schema validation in v2.

It's not quite as clear cut as the message failed validation though, a 400 could occur if:

* the movement ID is malformed (different to not found -- if the ID in the URL provided contains a non hex character it'll fail with a 400)
* the XML (or Json) is not valid XML (or Json).
* schema validation
* the office of destination is not a GB or XI office

Generally, the user did something wrong if they get a 400.

fmcgrath: Agreed. I wasn't happy with the original wording but I also felt that my proposed rewording did not go far enough.

**Suggestions**
_Option A_
The movement ID is malformed, or the JSON or XML is invalid, or the office of destination is not in Great Britain or Northern Ireland.

_Option B_
The request was incorrect or malformed.

Option B is terse but it's wording that we can reuse for other 400 errors.

drnaylor-oc: There is definately a wider question here on how we talk about errors on each endpoint -- I think what's happened is that we've taken the v2 endpoint descriptions from the v1 endpoint which wasn't clearly written in the first place.

I think a better way of looking at this will be what the post endpoints share in common error wise, and deciding the common wording from there...

derekspalmer (approved):

drnaylor-oc (approved):

resources/public/api/conf/2.0/application.raml
fmcgrath: The issues raised here have been discussed on Slack. It's been agreed that I will create a spike ticket ([CTCP-1544](https://jira.tools.tax.service.gov.uk/browse/CTCP-1544)) to explore some of these issues further.

PR #326, opened 2022-10-26
CTCP-1546: Fix IE170 audit type name

drnaylor-oc:


prashanthmamidi (dismissed):

test/v2/models/AuditTypeSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+
+package v2.models
+
+import org.scalatest.freespec.AnyFreeSpec
+import org.scalatest.matchers.must.Matchers
+
+class AuditTypeSpec extends AnyFreeSpec with Matchers {
+
+  "every AuditType must match the object name" - AuditType.values.foreach {
+    t =>
```
sonisn1988, line 25: Variable should be some meaningful name.

drnaylor-oc: I've updated it, not _completely_ sure it was required here as the context is clear, but it doesn't harm.

sonisn1988 (approved):

liandra-giacomine (approved):

PR #327, opened 2022-10-26
Allow departure messages to be returned in JSON dependent on accept header

liandra-giacomine:


app/v2/controllers/V2DeparturesController.scala
```diff
@@ -159,20 +161,42 @@ class V2DeparturesControllerImpl @Inject() (
...
     } yield declarationResult
 
-  def getMessage(departureId: MovementId, messageId: MessageId): Action[AnyContent] =
+  def getMessage(departureId: MovementId, messageId: MessageId, convertBodyToJson: Boolean): Action[AnyContent] = {
+
+    def checkAndConvertBodyToJson(messageSummary: MessageSummary)(implicit hc: HeaderCarrier): EitherT[Future, PresentationError, MessageSummary] =
+      if (!convertBodyToJson || messageSummary.body.isEmpty) EitherT.rightT(messageSummary)
+      else {
+        for {
+          jsonSource <- conversionService.jsonToXml(messageSummary.messageType, Source.single(ByteString(messageSummary.body.get))).asPresentation
```
drnaylor-oc, line 170: `xmlToJson`

app/routing/DeparturesRouter.scala
```diff
@@ -51,19 +51,28 @@ class DeparturesRouter @Inject() (
...
+    route {
+      case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE)          => routeToVersion2(false)
+      case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON_XML) => routeToVersion2(true)
+      case _ =>
+        runIfBound[V1DepartureId](
+          "departureId",
+          departureId,
+          boundDepartureId => runIfBound[V1MessageId]("messageId", messageId, v1DepartureMessages.getDepartureMessage(boundDepartureId, _))
+        )
+    }
```
drnaylor-oc, line 75: I think the router should only be concerned about routing to the correct version, and not more than that. The reason for this is that when we retire version 1 and delete all the v1 code, this router will likely go with it, and as it currently stands, you've got v2 specific logic in here.

I think the better solution is to pattern match on the regex `^application/vnd\.hrmc\.2\.0\+.+$` and let the v2 controller, rather than the router, dictate the return type -- that way when this router gets removed, the logic is still there.

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -159,20 +161,42 @@ class V2DeparturesControllerImpl @Inject() (
         .asPresentation
     } yield declarationResult
 
-  def getMessage(departureId: MovementId, messageId: MessageId): Action[AnyContent] =
+  def getMessage(departureId: MovementId, messageId: MessageId, convertBodyToJson: Boolean): Action[AnyContent] = {
```
drnaylor-oc, line 164: I don't like this boolean being here, because it's not clear what's going on, and if we for some reason want to add another format this will need to be changed.

Along with moving the selection logic to here, I'd create a sealed trait or abstract class like so:

```scala
sealed trait MessageFormat {
    def convertFromXML(message: String, conversionService: ConversionService): EitherT[Future, ConversionError(?), String] // or whatever
}

object XML {
    def convertFromXML(message: String, conversionService: ConversionService): EitherT[Future, ConversionError(?), String] == ??? // or whatever -- will be the identity
}

object Json {
 ...
}
``` 

and that way we can move some of this code away from here and make it more generic if we need it later.

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -159,20 +161,42 @@ class V2DeparturesControllerImpl @Inject() (
...
+    def checkAndConvertBodyToJson(messageSummary: MessageSummary)(implicit hc: HeaderCarrier): EitherT[Future, PresentationError, MessageSummary] =
+      if (!convertBodyToJson || messageSummary.body.isEmpty) EitherT.rightT(messageSummary)
+      else {
+        for {
+          jsonSource <- conversionService.jsonToXml(messageSummary.messageType, Source.single(ByteString(messageSummary.body.get))).asPresentation
+          jsonBody <- EitherT {
+            jsonSource
+              .fold("")(
+                (acc, byteStr) => acc + byteStr.utf8String
+              )
```
drnaylor-oc, line 175: It'll be more effcient to concatenate `ByteStrings`, and we could use reduce, so we could do

```
.reduce(_ ++ _).map(_.utf8String)
```

instead.

app/v2/controllers/V2DeparturesController.scala
drnaylor-oc: (note that names and such aren't final, more interested in getting the idea across with this!)

app/routing/DeparturesRouter.scala
drnaylor-oc: Also, additionally, if I set the accept header to be `application/vnd.hmrc.2.0+xml`, which may become a valid option in the future, this will redirect to the v1 route, though I think that's a problem with all routing we have and we might need to look into that...

liandra-giacomine: @drnaylor-oc yep agreed, good point

app/routing/DeparturesRouter.scala
```diff
@@ -51,20 +51,28 @@ class DeparturesRouter @Inject() (
...
-      )
-    case _ =>
-      runIfBound[V1DepartureId](
-        "departureId",
-        departureId,
-        boundDepartureId => runIfBound[V1MessageId]("messageId", messageId, v1DepartureMessages.getDepartureMessage(boundDepartureId, _))
-      )
-  }
+  def isValidVersionTwoAcceptHeader(headerValue: String) = """^application\/vnd\.hmrc\.2\.0\+.+$""".r.pattern.matcher(headerValue).matches
+
```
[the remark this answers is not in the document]
RobertLaverick: Might be cleaner and stricter (if we extract the accept header elsewhere) to use something like this.

``` case Some(VERSION_2_ACCEPT_HEADER_VALUE) | Some(VERSION_2_ACCEPT_HEADER_VALUE_JSON_XML) =>
.... v2 logic
case Some(acceptHeaderValue) if(v2Regex) => Reject
case _ => v1 logic

File: test/v2/controllers/V2DeparturesControllerSpec.scala [hunk 6]
```diff
@@ -791,61 +791,77 @@ class V2DeparturesControllerSpec
 
   }
 
-  "for retrieving a single message" - {
+  Seq(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE, VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON_XML).foreach {
+    acceptHeaderValue =>
```
RobertLaverick, line 795: You're only testing the accepted header, there are no checks to confirm the 406 behaviour

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -159,19 +159,30 @@ class V2DeparturesControllerImpl @Inject() (
...
+          .flatMap {
+            messageSummary =>
+              acceptHeaderValue match {
+                case value if value == VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON_XML && messageSummary.body.isDefined =>
+                  XmlMessage
+                    .convertToJson(messageSummary.messageType, messageSummary.body.get, conversionService)
+                    .map(
+                      jsonBody => messageSummary.copy(body = Some(jsonBody))
+                    )
+                case _ => EitherT.rightT[Future, PresentationError](messageSummary)
```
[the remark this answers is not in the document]
RobertLaverick: In addition, I'm not sure you need the XmlMessage object, since it's basically just a wrapper around the conversion call.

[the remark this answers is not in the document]
RobertLaverick: Also, I think you've got the headers mixed up, if JSON+XML is meant to be XML wrapped in Json, then you should be using the existing header and leaving it alone on the new one? (I think?)

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -159,19 +159,30 @@ class V2DeparturesControllerImpl @Inject() (
         .asPresentation
     } yield declarationResult
 
-  def getMessage(departureId: MovementId, messageId: MessageId): Action[AnyContent] =
+  def getMessage(departureId: MovementId, messageId: MessageId, acceptHeaderValue: String): Action[AnyContent] =
```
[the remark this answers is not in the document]
RobertLaverick: Also, when we retire P4, the DeparturesRouter won't (or probably won't) exist any more unless we rapidly start a v3 api so we'd have to extract the header after the transition anyway.

app/routing/DeparturesRouter.scala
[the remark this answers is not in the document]
drnaylor-oc: >  It feels like this should live in the VersionedRouting object (or at least with the two accept headers).
>
> Wouldn't this technically pass for "application/vnd.hmrc.2.0+edifact"? which we don't support?

This change is actually one I asked for.

I actually think that's okay in the router. It's not for the router to decide what formats we accept, it's just to decide what version the call is.

We should be enforcing the correct accept formats in the specific versioned controllers, not in the router. The router should not care about formats.

app/v2/controllers/V2DeparturesController.scala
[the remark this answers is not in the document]
drnaylor-oc: This is also where we should be enforcing the formats, so I think we should match on the XML+JSON header and the JSON header explicity, and reject all other accept headers with a 406.

app/v2/controllers/MessageFormat.scala
```diff
@@ -0,0 +1,61 @@
...
+    source
+      .reduce(_ ++ _)
+      .map(_.utf8String)
+      .runWith(Sink.head[String])
+      .map(Right[PresentationError, String])
+      .recover {
+        case NonFatal(ex) => Left[PresentationError, String](PresentationError.internalServiceError(cause = Some(ex)))
+      }
+  }
+}
```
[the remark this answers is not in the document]
drnaylor-oc: I agree -- this is something that may have use outside of the controller!

app/routing/DeparturesRouter.scala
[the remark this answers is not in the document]
drnaylor-oc: (I think I need to create a ticket for enforcing the right formats on the various controllers to ensure the format is correct...)

app/v2/controllers/MessageFormat.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  def convertToJson(messageType: MessageType, body: String, conversionService: ConversionService)(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext,
+    materializer: Materializer
+  ): EitherT[Future, PresentationError, String] =
+    for {
+      jsonSource <- conversionService.xmlToJson(messageType, Source.single(ByteString(body))).asPresentation
+      jsonBody   <- convertSourceToString(jsonSource)
+    } yield jsonBody
```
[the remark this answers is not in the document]
drnaylor-oc: There will be some value in extracting this because this is going to be something needed in the arrivals controller too... but at that point it'll probably be worth making this a trait that will contain multiple common messages.

For now though? I'd probably just stick this in the controller itself, or if there is already a common trait for controller logic, use that.

app/v2/controllers/V2DeparturesController.scala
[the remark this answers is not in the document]
drnaylor-oc: actually, I see this happens quite late on -- we should be performing that rejection earlier.

app/v2/controllers/V2DeparturesController.scala
```diff
@@ -159,21 +163,42 @@ class V2DeparturesControllerImpl @Inject() (
...
+        }
     }
 
+  def getMessageJsonWrappedXml(movementId: MovementId, messageId: MessageId)(implicit hc: HeaderCarrier, request: AuthenticatedRequest[_]): Future[Result] =
+    (for {
+      messageSummary <- departuresService.getMessage(request.eoriNumber, movementId, messageId).asPresentation
+    } yield messageSummary).fold(
+      presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
+      response => Ok(Json.toJson(HateoasDepartureMessageResponse(movementId, messageId, response)))
+    )
```
drnaylor-oc, line 183: This is slightly nitpicky, but the bracket surrouding the for comprehension is redundant. Normally I wouldn't really bat an eyelid at this, but the way the brackets are laid out, I spent 15 seconds trying to work out how this worked without one more bracket -- the bracket that matches the fold is on line 182!

That said, we don't need a for comprehension for this:

```suggestion
    departuresService
        .getMessage(request.eoriNumber, movementId, messageId)
        .asPresentation
        .fold(
            presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
            response => Ok(Json.toJson(HateoasDepartureMessageResponse(movementId, messageId, response))
        )
```

drnaylor-oc: (What you've written is fine, actually, but the bracket irked me, apparently!)

app/v2/controllers/AcceptHeaderRouting.scala
```diff
@@ -0,0 +1,60 @@
...
+import play.api.mvc.AnyContent
+import play.api.mvc.BaseController
+import play.api.mvc.Request
+import play.api.mvc.Result
+import v2.controllers.stream.StreamingParsers
+import v2.models.errors.PresentationError
+
+import scala.concurrent.Future
+
+trait AcceptHeaderRouting {
```
drnaylor-oc, line 35: We should have a separate test suite for this.

app/v2/utils/MessageFormat.scala
```diff
@@ -0,0 +1,46 @@
...
+  def convertSourceToString(
+    source: Source[ByteString, _]
+  )(implicit ec: ExecutionContext, mat: Materializer): EitherT[Future, MessageFormatError, String] = EitherT {
+    source
+      .reduce(_ ++ _)
+      .map(_.utf8String)
+      .runWith(Sink.head[String])
+      .map(Right[MessageFormatError, String])
+      .recover {
+        case NonFatal(ex) => Left[MessageFormatError, String](MessageFormatError.UnexpectedError(Some(ex)))
```
RobertLaverick, line 43: Not sure about the name, code's fine, just MessageFormat seems a bit weird for something that's essentially a tool to realise or force a stream to complete into a string. StreamComplete? CompleteSource? ResolveSource?

drnaylor-oc: StreamingUtils?

drnaylor-oc:
It failed the build -- my suggestion was missing a bracket!

app/v2/controllers/AcceptHeaderRouting.scala
```diff
@@ -0,0 +1,53 @@
...
+                    headerValue => s"Accept header $headerValue is not supported!"
+                  )
+                  .getOrElse("Accept header is required!")
+              )
+            )
+          )
+        )
+      }
+
+}
```
RobertLaverick, line 53: This would work really well as an ActionRefiner (I think) to both provide a new request type with the accept header stored in it as a case object, and provide the rejection case.

app/v2/controllers/ResponseFormatter.scala
```diff
@@ -0,0 +1,56 @@
...
+      case (VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON, MessageSummary(_, _, messageType, Some(body))) =>
+        for {
+          jsonSource <- conversionService.xmlToJson(messageType, Source.single(ByteString(body))).asPresentation
+          jsonBody   <- StreamingUtils.convertSourceToString(jsonSource).asPresentation
+        } yield HateoasDepartureMessageResponse(movementId, messageId, messageSummary.copy(body = Some(jsonBody)))
+      case _ =>
+        EitherT.rightT(HateoasDepartureMessageResponse(movementId, messageId, messageSummary))
+    }
+
+}
```
RobertLaverick, line 56: I'm not sure why this is an object rather than an injected service which would also be able to inject the conversionService into here via the constructor. Also, you don't need to go so far as to produce the full response (my mistake, I worded myself poorly I think) - you just need to replace the body field of the message summary and return that.

drnaylor-oc:
Sorry in advance if this throws a spanner in the works...

What does the Json summary look like, specifically the body field? Does it look like this?

```json
{
   ...,
    "body": "{ \"n1.CC015C\": { \"messageType\": \"CC015C\", ... } }"
}
```

or 

```json
{
   ...,
    "body": { 
        "n1.CC015C": { 
            "messageType": "CC015C",
            ... 
        } 
    }
}
```

i.e., the message Json is a full Json object, not a stringified version? I imagine it's the former, but should it be the latter?

app/v2/models/responses/hateoas/HateoasDepartureMessageResponse.scala
```diff
@@ -17,14 +17,17 @@
...
 import play.api.libs.json.Json
+import routing.VersionedRouting
 import v2.models.MessageId
 import v2.models.MovementId
 import v2.models.responses.MessageSummary
 
 object HateoasDepartureMessageResponse extends HateoasResponse {
 
-  def apply(departureId: MovementId, messageId: MessageId, messageSummary: MessageSummary): JsObject =
+  def apply[A](departureId: MovementId, messageId: MessageId, messageSummary: MessageSummary, acceptHeader: String): JsObject =
```
drnaylor-oc, line 30: We don't need this generic any more.

app/v2/models/Payload.scala
```diff
@@ -0,0 +1,30 @@
...
+import play.api.libs.json.Json
+
+object Payload {
+  implicit lazy val payloadFormat: Format[Payload] = Json.valueFormat[Payload]
+}
+
+case class Payload(value: String) {
+  lazy val asJson = Json.parse(value)
+  lazy val asXml  = JsString(value)
+}
```
drnaylor-oc, line 30: Ah, okay, this isn't quite what we meant yesterday, rather:

```scala
abstract class Payload(val value: String) {
    def asJson: JsValue
}

case class XMLPayload(value: String) extends Payload(value) {
    lazy val asJson: JsValue = JsString(value)
}

case class JsonPayload(value: String) extends Payload(value) {
    lazy val asJson: JsValue = Json.parse(value)
}
```

Then in the message summary reads, we'll need to specifically choose `XMLPayload` -- however, in the `MessageSummary` itself, the body must be of type `Payload`. For the cases we need to change from XML to Json, we swap the XMLPayload for the JsonPayload

app/routing/DeparturesRouter.scala
```diff
@@ -47,27 +47,35 @@ class DeparturesRouter @Inject() (
...
+  def getMessage(departureId: String, messageId: String): Action[Source[ByteString, _]] =
+    route {
+      case Some(acceptHeaderValue) if isValidVersionTwoAcceptHeader(acceptHeaderValue) =>
```
drnaylor-oc, line 58: Don't need you to change this, becauase this is perfectly fine, but you can bake the pattern match into the case statement like this.

```scala
val VERSION_2_ACCEPT_HEADER_PATTERN = """^application\/vnd\.hmrc\.2\.0\+.+$""".r

def getMessage(departureId: String, messageId: String): Action[Source[ByteString, _]] =
    route {
        case Some(VERSION_2_ACCEPT_HEADER_PATTERN()) => // v2
        case _ => // v1
    }
```

It's stylistic, but something to consider for the future.

app/v2/controllers/request/BodyReplaceableRequests.scala
```diff
@@ -26,4 +27,5 @@ abstract class BodyReplaceableRequest[+R[_], B](request: Request[B]) extends Wra
 
 case class AuthenticatedRequest[A](eoriNumber: EORINumber, request: Request[A]) extends BodyReplaceableRequest[AuthenticatedRequest, A](request) {
   override def replaceBody(body: A): AuthenticatedRequest[A] = AuthenticatedRequest(eoriNumber, request.withBody(body))
+  val acceptHeader                                           = request.headers.get(HeaderNames.ACCEPT)
```
drnaylor-oc, line 30: Is this needed now? I suspect this is a bit of a hangover from an earlier iteration -- the only place that this is used as far as I can see is once in the controller -- we might as well just get it via `request.headers.get(HeaderNames.ACCEPT)` there.

If you feel it is needed, then fair enough, just not convinced this is!

drnaylor-oc (dismissed):

PR #329, opened 2022-11-04
[CTCP-1452] added implementation for Get arrivals by EORI API

prashanthmamidi:


app/routing/ArrivalsRouter.scala
```diff
@@ -42,4 +44,10 @@ class ArrivalsRouter @Inject() (
       case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE) => v2Arrivals.createArrivalNotification()
```
liandra-giacomine, line 44: VERSION_2_ACCEPT_HEADER_VALUE looks outdated, should be VERSION_2_ACCEPT_HEADER_VALUE_JSON - need to merge with main

app/routing/ArrivalsRouter.scala
```diff
@@ -42,4 +44,10 @@ class ArrivalsRouter @Inject() (
...
       case _                                                    => v1Arrivals.createArrivalNotification()
     }
+
+  def getArrivalsForEori(updatedSince: Option[OffsetDateTime] = None): Action[Source[ByteString, _]] = route {
+    case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE) => v2Arrivals.getArrivalsForEori(updatedSince)
```
sonisn1988, line 49: Test cases for ArrivalsRouter seems missing for this method implementation in ArrivalsRouterSpec

app/routing/ArrivalsRouter.scala
```diff
@@ -42,4 +44,10 @@ class ArrivalsRouter @Inject() (
       case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE) => v2Arrivals.createArrivalNotification()
       case _                                                    => v1Arrivals.createArrivalNotification()
     }
+
+  def getArrivalsForEori(updatedSince: Option[OffsetDateTime] = None): Action[Source[ByteString, _]] = route {
```
sonisn1988, line 48: Can not see any usage of updatedSince optional query param in code also not used in transit-movements backend connector api. any specific reason to use this optional query param?

app/v2/controllers/V2ArrivalsController.scala
```diff
@@ -132,6 +137,20 @@ class V2ArrivalsControllerImpl @Inject() (
         } yield arrivalResult
     }
 
+  def getArrivalsForEori(updatedSince: Option[OffsetDateTime]): Action[AnyContent] =
```
sonisn1988, line 140: can not see any usage of updatedSince:?

conf/app.routes
```diff
@@ -1,4 +1,4 @@
-GET        /movements/arrivals                                      controllers.ArrivalMovementController.getArrivalsForEori(updatedSince: Option[OffsetDateTime] ?= None)
+GET        /movements/arrivals                                      routing.ArrivalsRouter.getArrivalsForEori(updatedSince: Option[OffsetDateTime] ?= None)
```
sonisn1988, line 1: Again con not see any usage of updatedSince: Option[OffsetDateTime] in code

app/v2/controllers/V2ArrivalsController.scala
drnaylor-oc: Interstingly, this isn't on the equivalent method on the Departures controller either, so I wonder if it's supported in the backend (transit-movements). We should probably look to fix that via another ticket if it isn't. If it is supported in the backend, then it should be supported here, and a ticket should be raised for the departures side of things.

test/v2/base/CommonGenerators.scala
```diff
@@ -74,4 +75,15 @@ trait CommonGenerators {
     Gen.oneOf(MovementType.values)
   }
 
+  implicit lazy val genMovementResponse: Arbitrary[MovementResponse] = Arbitrary {
```
sonisn1988, line 78: Name should be start with arbitrary. as this is arbitrary not gen. Like arbitraryMovementResponse

test/v2/models/responses/hateoas/HateosMovementResponseSpec.scala
```diff
@@ -24,20 +24,20 @@ import play.api.libs.json.Json
...
-import v2.models.responses.DepartureResponse
+import v2.models.responses.MovementResponse
 
 import java.time.OffsetDateTime
 import java.time.ZoneOffset
 
-class HateosDepartureResponseSpec extends AnyFreeSpec with Matchers with ScalaCheckDrivenPropertyChecks {
+class HateosMovementResponseSpec extends AnyFreeSpec with Matchers with ScalaCheckDrivenPropertyChecks {
 
   private val hexId    = Gen.listOfN(16, Gen.hexChar).map(_.mkString.toLowerCase)
```
sonisn1988, line 34: This can be moved to CommonGenerators

liandra-giacomine: Yes, you can use arbitraryMovementId in CommonGenerators, which generates the hexId and creates the object

test/v2/controllers/V2ArrivalsControllerSpec.scala
```diff
@@ -675,4 +682,95 @@ class V2ArrivalsControllerSpec
...
+  "GET  /traders/:EORI/movements/arrivals" - {
+    "should return ok with json body for arrivals" in {
+
+      val enrolmentEORINumber = arbitrary[EORINumber].sample.value
+      val dateTime            = OffsetDateTime.of(2022, 8, 4, 11, 34, 42, 0, ZoneOffset.UTC)
+
+      val movementResponse1 = MovementResponse(
+        _id = arbitrary[MovementId].sample.value,
+        enrollmentEORINumber = enrolmentEORINumber,
+        movementEORINumber = arbitrary[EORINumber].sample.value,
```
sonisn1988, line 694: can use the Seq of Arbitrary[MovementResponse] instead of creating two new instances here. Not sure.
@liandra-giacomine / @drnaylor-oc any suggestions

test/v2/services/ArrivalsServiceSpec.scala
```diff
@@ -92,4 +100,45 @@ class ArrivalsServiceSpec
...
+        when(mockConnector.getArrivalsForEori(EORINumber("1")))
+          .thenReturn(Future.successful(expected))
+
+        val result = sut.getArrivalsForEori(EORINumber("1"))
+        whenReady(result.value) {
+          _ mustBe Right(expected)
+        }
+    }
+
+    "when a arrival is not found, should return a Left with an ArrivalNotFound" in {
```
sonisn1988, line 121: This should be ArrivalsNotFound ?

test/v2/services/ArrivalsServiceSpec.scala
```diff
@@ -92,4 +100,45 @@ class ArrivalsServiceSpec
     }
   }
 
+  "Getting a list of Arrivals (Movement) by EORI" - {
+
+    val dateTime = Gen.option(arbitrary[OffsetDateTime])
+
+    val now = OffsetDateTime.now(ZoneOffset.UTC)
+
+    "when a departure (movement) is found, should return a Right" in forAll(Gen.listOfN(3, arbitrary[MovementResponse])) {
```
sonisn1988, line 109: this should be arrival(movement)?

test/v2/services/ArrivalsServiceSpec.scala
```diff
@@ -92,4 +100,45 @@ class ArrivalsServiceSpec
     }
   }
 
+  "Getting a list of Arrivals (Movement) by EORI" - {
+
+    val dateTime = Gen.option(arbitrary[OffsetDateTime])
```
sonisn1988, line 105: not using datetime with this api. Should be removed.

test/v2/controllers/V2ArrivalsControllerSpec.scala
```diff
@@ -675,4 +682,95 @@ class V2ArrivalsControllerSpec
...
+
+      status(result) mustBe OK
+      contentAsJson(result) mustBe Json.toJson(
+        HateoasArrivalIdsResponse(
+          movementResponses
+        )
+      )
+    }
+
+    "should return arrival not found if persistence service returns 404" in {
```
sonisn1988, line 730: should return arrivals not found

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -920,9 +920,108 @@ class PersistenceConnectorSpec
     }
   }
 
-  private def generateDepartureResponse(departureId: MovementId) =
-    DepartureResponse(
-      _id = departureId,
+  "GET /traders/movements/arrivals" - {
```
sonisn1988, line 923: endpoint with this should be GET /traders/${eoriNumber.value}/movements/arrivals?

test/v2/controllers/V2ArrivalsControllerSpec.scala
liandra-giacomine: Could use arbitraryMovementResponse from CommonGenerators, then use the .copy method to just update the relevant parameters

it/v2/connectors/PersistenceConnectorSpec.scala
liandra-giacomine: No, eoriNumber is passed as a query parameter in the api

sonisn1988: yes agree. But this is connector and  basically written to call the transit-movements api  endpoint /traders/:EORI/movements/arrivals . I am fine with both. Just asked if change is required.

app/v2/controllers/V2ArrivalsController.scala
```diff
@@ -45,11 +48,13 @@ import v2.models.responses.ArrivalResponse
 import v2.models.responses.hateoas._
 import v2.services._
 
+import java.time.OffsetDateTime
 import scala.concurrent.Future
 
 @ImplementedBy(classOf[V2ArrivalsControllerImpl])
 trait V2ArrivalsController {
```
sonisn1988, line 55: Add api definition documentation in raml for version 2.0 implementation

it/v2/connectors/PersistenceConnectorSpec.scala
liandra-giacomine: Oh yeah sorry you're right

app/v2/controllers/V2ArrivalsController.scala
prashanthmamidi: Yes, by default we sort by descending order in _transit-movements_     
```
val aggregates = Seq(
      Aggregates.filter(selector),
      Aggregates.sort(descending("updated")),
      Aggregates.project(projection)
    )
```

test/v2/models/responses/hateoas/HateosMovementResponseSpec.scala
prashanthmamidi: yes, we can use arbitraryMovementId

conf/app.routes
prashanthmamidi: as mentioned above, it will be dealt in separate ticket

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -921,9 +921,108 @@ class PersistenceConnectorSpec
...
+          case UpstreamErrorResponse(_, status, _, _) => status
+          case thr                                    => fail(s"Expected an UpstreamErrorResponse with a 500, got $thr")
+        }
+
+      whenReady(result) {
+        _ mustBe INTERNAL_SERVER_ERROR
+      }
+    }
+
+  }
```
sonisn1988, line 1021: please add test cases to cover 404 not found

test/routing/ArrivalsRouterSpec.scala
```diff
@@ -91,4 +92,34 @@ class ArrivalsRouterSpec extends AnyFreeSpec with Matchers with OptionValues wit
     }
   }
 
+  "when fetching arrivals by EORI " - {
+
+    "with accept header set to application/vnd.hmrc.2.0+json (version two)" - {
+      val arrivalsHeaders = FakeHeaders(
+        Seq(HeaderNames.ACCEPT -> VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON, HeaderNames.CONTENT_TYPE -> "application/xml")
```
sonisn1988, line 99: content type is not required to pass in case of get ?

test/v2/controllers/V2ArrivalsControllerSpec.scala
```diff
@@ -675,4 +682,80 @@ class V2ArrivalsControllerSpec
     }
   }
 
+  "GET  /traders/:EORI/movements/arrivals" - {
```
sonisn1988, line 685: endpoint looks incorrect should be GET  /traders/movements/arrivals? for controller

test/v2/controllers/V2ArrivalsControllerSpec.scala
```diff
@@ -675,4 +682,80 @@ class V2ArrivalsControllerSpec
...
+    "should return ok with json body for arrivals" in {
+
```
sonisn1988, line 687: we can use forAll(Gen.listOfN(arbitrary[MovementResponse]) something like that with this?

it/v2/connectors/PersistenceConnectorSpec.scala
liandra-giacomine: 404 is handled by the service

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -921,9 +921,108 @@ class PersistenceConnectorSpec
...
+                  )
+                )
+              )
+          )
+      )
+
+      val result = persistenceConnector
+        .getArrivalsForEori(eori)
+        .map(
+          _ => fail("This should have failed with an UpstreamErrorResponse, but it succeeded")
```
sonisn1988, line 1009: This working could be improved . Something like below

          .map(
            _ => fail("This should not succeed")
          )
          .recover {
            case UpstreamErrorResponse(_, INTERNAL_SERVER_ERROR, _, _) =>
          }

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -921,9 +921,108 @@ class PersistenceConnectorSpec
...
+      )
+
+      implicit val hc = HeaderCarrier()
+      val result      = persistenceConnector.getArrivalsForEori(eori)
+      whenReady(result) {
+        _ mustBe messageResponse
+      }
+    }
+
+    "on incorrect Json, return an error" in {
```
sonisn1988, line 957: I am confused with this test case. is this valid scenario to test ? @drnaylor-oc

it/v2/connectors/PersistenceConnectorSpec.scala
drnaylor-oc: The wording is perfectly fine -- in fact, what has been written here is better because it explains exactly what happened.

it/v2/connectors/PersistenceConnectorSpec.scala
drnaylor-oc: It's testing if an upstream failure occurs, but in a non-standard way. It's fine.

it/v2/connectors/PersistenceConnectorSpec.scala
sonisn1988: I think we can handle 404 here as well. Because from transit-movements we are handing both type of error 500 and 404. I am ok if not required

it/v2/connectors/PersistenceConnectorSpec.scala
sonisn1988: No problem. but this test case look like is written to cover 500 Internal server error. so how it can succeed?  the wording says but it succeeded. Still I am confused.

drnaylor-oc: The `map` is only executed when the call succeeds. Since we're checking that we handle the _failure_ case correctly, then if we end up in the `map` method, we didn't get an error back and so the call succeeded -- so this should fail the test.

sonisn1988: Ok fine. I misunderstood this.

it/v2/connectors/PersistenceConnectorSpec.scala
liandra-giacomine: @drnaylor-oc To add to that, I believe it's testing if the body returned can't be deserialised into the expected type

test/v2/controllers/V2ArrivalsControllerSpec.scala
prashanthmamidi: it should be fine, we can keep the list as is in seq[], can't see adding much value to this

drnaylor-oc (approved):

sonisn1988 (approved):

PR #330, opened 2022-11-04
CTCP-1523: Update API Definition for additional accept header value

drnaylor-oc:
There is one thing I'm a little bothered about, which is that the response examples don't include the description. We get this (via the API definition preview RAML function):

image

With the switch to OAS this might not be a problem, but for now, given we're stuck with RAML, we need to consider if we're happy with this or not.

@fmcgrath I'm sure I could word things better!

### DIFF
--- a/resources/public/api/conf/2.0/application.raml
+++ b/resources/public/api/conf/2.0/application.raml
@@ -49,6 +49,16 @@ traits:
         type: string
         required: true
         example: application/vnd.hmrc.2.0+json
+  acceptSingleMessageGetHeader:
+    headers:
+      Accept:
+        description: |
+          Specifies the response format, which must be JSON or JSON+XML, and the [version](/api-documentation/docs/reference-guide#versioning) of the API to be used.
+          
+          Acceptable values are "application/vnd.hmrc.2.0+json" or "application/vnd.hmrc.2.0+json-xml"
+        type: string
+        required: true
+        example: application/vnd.hmrc.2.0+json-xml
   acceptHeaderInvalid:
     responses:
       406:
@@ -130,7 +140,7 @@ traits:
         Movements older than 31 days and their related messages are archived and cannot be returned by the API.
 
         A successful response will include a Movement Reference Number (MRN) only after it has been allocated to a movement.
-
+        
         **Note:** You can view a transit movement only if your Economic Operators Registration and Identification (EORI) number matches the EORI number that created the transit declaration for that movement. Your EORI number is part of the bearer token that you use for API authentication.
       is:
         - acceptJsonHeader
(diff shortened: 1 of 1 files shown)


resources/public/api/conf/2.0/application.raml
```diff
@@ -336,16 +346,41 @@ traits:
               For example, this could include ‘Declaration rejected’ or Movement Reference Number (MRN) allocated messages.
 
               Movements older than 31 days and their related messages are archived and cannot be returned by the API.
-
+              
+              The response can be returned in one of the following formats based on the provided accept header:
+        
+              * `application/vnd.hmrc.2.0+json-xml`: Json wrapped XML, where the message is provided as an XML string, sent as part of a Json payload containing other metadata
+              * `application/vnd.hmrc.2.0+json`: Json, where the message is provided as a sub document sent as part of a Json payload containing other metadata
```
drnaylor-oc, line 353: We've written Json as JSON elsewhere, so I should update this to be JSON... but I'll do this after comments.

resources/public/api/conf/2.0/application.raml
```diff
@@ -49,6 +49,16 @@ traits:
         type: string
         required: true
         example: application/vnd.hmrc.2.0+json
+  acceptSingleMessageGetHeader:
+    headers:
+      Accept:
+        description: |
+          Specifies the response format, which must be JSON or JSON+XML, and the [version](/api-documentation/docs/reference-guide#versioning) of the API to be used.
+          
+          Acceptable values are "application/vnd.hmrc.2.0+json" or "application/vnd.hmrc.2.0+json-xml"
```
fmcgrath, line 58: Should back ticks be used?
`application/vnd.hmrc.2.0+json`
`application/vnd.hmrc.2.0+json-xml`

End of second sentence needs a full stop.

resources/public/api/conf/2.0/application.raml
fmcgrath: I agree with @drnaylor-oc and go further.

Suggestion:
 The response can be returned in one of the following formats based on the provided `Accept` header:

- `application/vnd.hmrc.2.0+json-xml`: JSON-wrapped XML, where the message is provided as an XML string sent as part of JSON payload containing other metadata
- `application/vnd.hmrc.2.0+json`: JSON, where the message is provided as a sub-document sent as part of a JSON payload containing other metadata

fmcgrath (changes requested):
No showstoppers but some formatting and wording problems to be resolved.

resources/public/api/conf/2.0/application.raml
```diff
@@ -347,10 +347,10 @@ traits:
 
               Movements older than 31 days and their related messages are archived and cannot be returned by the API.
               
-              The response can be returned in one of the following formats based on the provided accept header:
+              The response can be returned in one of the following formats based on the provided `Accept` header:
         
-              * `application/vnd.hmrc.2.0+json-xml`: Json wrapped XML, where the message is provided as an XML string, sent as part of a Json payload containing other metadata
-              * `application/vnd.hmrc.2.0+json`: Json, where the message is provided as a sub document sent as part of a Json payload containing other metadata
+              * `application/vnd.hmrc.2.0+json-xml`: JSON-wrapped XML, where the message is provided as an XML string sent as part of JSON payload containing other metadata.
+              * `application/vnd.hmrc.2.0+json`: LSON, where the message is provided as a sub-document sent as part of a JSON payload containing other metadata.
```
fmcgrath, line 354: Typo: `LSON` instead of `JSON`.

fmcgrath (commented):
Typo.

resources/public/api/conf/2.0/application.raml
drnaylor-oc: _facepalm_

fmcgrath (approved):
Looks good.

PR #332, opened 2022-11-07
CTCP 1463: Get arrival by arrival id

liandra-giacomine:


app/controllers/ArrivalMessagesController.scala
```diff
@@ -38,7 +38,7 @@ import play.api.mvc.AnyContent
 import play.api.mvc.ControllerComponents
 import uk.gov.hmrc.http.HttpErrorFunctions
 import uk.gov.hmrc.play.bootstrap.backend.controller.BackendController
-import utils.CallOps._
+import v2.utils.CallOps._
```
sonisn1988, line 41: should use utils.CallOps._ for v1

app/controllers/ArrivalMovementController.scala
```diff
@@ -38,7 +38,7 @@ import play.api.mvc.AnyContent
 import play.api.mvc.ControllerComponents
 import uk.gov.hmrc.http.HttpErrorFunctions
 import uk.gov.hmrc.play.bootstrap.backend.controller.BackendController
-import utils.CallOps._
+import v2.utils.CallOps._
```
sonisn1988, line 41: should use utils.CallOps._ for v1

app/controllers/DepartureMessagesController.scala
```diff
@@ -41,7 +41,7 @@ import play.api.mvc.AnyContent
 import play.api.mvc.ControllerComponents
 import uk.gov.hmrc.http.HttpErrorFunctions
 import uk.gov.hmrc.play.bootstrap.backend.controller.BackendController
-import utils.CallOps._
+import v2.utils.CallOps._
```
sonisn1988, line 44: should use utils.CallOps._ for v1

app/controllers/DeparturesController.scala
```diff
@@ -37,7 +37,7 @@ import play.api.mvc.AnyContent
 import play.api.mvc.ControllerComponents
 import uk.gov.hmrc.http.HttpErrorFunctions
 import uk.gov.hmrc.play.bootstrap.backend.controller.BackendController
-import utils.CallOps._
+import v2.utils.CallOps._
```
sonisn1988, line 40: should use utils.CallOps._ for v1

app/controllers/ArrivalMessagesController.scala
liandra-giacomine: @sonisn1988 I moved it v2 because we now use it in HateosResponse rather than using the if statement 
https://github.com/hmrc/common-transit-convention-traders/pull/332/files#diff-c531ed67eb289711a1e6fa6155247b9095162c1e78f0b49fcf37169f2d43035dL36

When we delete V1 components we will want this to remain available

app/v2/models/responses/hateoas/HateoasMovementResponse.scala
```diff
@@ -0,0 +1,46 @@
...
+import play.api.libs.json.Json
+import v2.models.MovementId
+import v2.models.MovementType
+import v2.models.responses.MovementResponse
+
+object HateoasMovementResponse extends HateoasResponse {
+
+  def apply(movementId: MovementId, movementResponse: MovementResponse, movementType: MovementType): JsObject = {
+    val selfLink     = if (movementType == MovementType.Arrival) arrivalUri(movementId) else departureUri(movementId)
+    val messagesLink = if (movementType == MovementType.Arrival) arrivalMessageIdsUri(movementId) else messageIdsUri(movementId, None)
```
sonisn1988, line 29: messageIdsUri should changed to departureMessageIdsUri?

it/connectors/ArrivalConnectorSpec.scala
```diff
@@ -205,7 +205,7 @@ class ArrivalConnectorSpec
       val connector = app.injector.instanceOf[ArrivalConnector]
       val arrival = Arrival(
         ArrivalId(1),
-        routes.ArrivalMovementController.getArrival(ArrivalId(1)).urlWithContext,
+        routing.routes.ArrivalsRouter.getArrival(ArrivalId(1).toString).urlWithContext,
```
sonisn1988, line 208: we can pass directly string here "1" instead of ArrivalId(1).toString? because in ArrivalRouter we are now passing string directly

it/connectors/ArrivalMessageConnectorSpec.scala
```diff
@@ -86,7 +86,7 @@ class ArrivalMessageConnectorSpec
       val connector = app.injector.instanceOf[ArrivalMessageConnector]
       val arrival = Arrival(
         ArrivalId(1),
-        routes.ArrivalMovementController.getArrival(ArrivalId(1)).urlWithContext,
+        routing.routes.ArrivalsRouter.getArrival(ArrivalId(1).toString).urlWithContext,
```
sonisn1988, line 89: same here

test/controllers/ArrivalMessagesControllerSpec.scala
```diff
@@ -92,7 +92,7 @@ class ArrivalMessagesControllerSpec
 
   val sourceArrival = ArrivalWithMessages(
     ArrivalId(123),
-    routes.ArrivalMovementController.getArrival(ArrivalId(123)).urlWithContext,
+    routing.routes.ArrivalsRouter.getArrival(ArrivalId(123).toString).urlWithContext,
```
sonisn1988, line 95: same here

test/controllers/ArrivalMovementControllerSpec.scala
```diff
@@ -84,7 +84,7 @@ class ArrivalMovementControllerSpec
 
   val sourceArrival = Arrival(
     ArrivalId(123),
-    routes.ArrivalMovementController.getArrival(ArrivalId(123)).urlWithContext,
+    routing.routes.ArrivalsRouter.getArrival(ArrivalId(123).toString).urlWithContext,
```
sonisn1988, line 87: same here string

test/routing/ArrivalsRouterSpec.scala
```diff
@@ -48,49 +48,56 @@ class ArrivalsRouterSpec extends AnyFreeSpec with Matchers with OptionValues wit
...
+        )
 
-      "must route to the v2 controller and return Accepted when successful" in {
+        "must route to the v2 controller and return Accepted when successful" in {
 
-        val request =
-          FakeRequest(method = "POST", uri = routes.ArrivalsRouter.createArrivalNotification().url, body = <test></test>, headers = arrivalsHeaders)
-        val result = call(sut.createArrivalNotification(), request)
+          val request =
+            FakeRequest(method = "POST", uri = endpoint.url, body = <test></test>, headers = arrivalsHeaders)
```
sonisn1988, line 68: should not use endpoint.method here ?

app/v2/connectors/PersistenceConnector.scala
```diff
@@ -272,4 +277,22 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
...
 
+  override def getArrival(eori: EORINumber, movementId: MovementId)(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): Future[MovementResponse] = {
+    val url = appConfig.movementsUrl.withPath(movementsGetArrival(eori, movementId))
+
+    httpClientV2
+      .get(url"$url")
+      .addHeaders(HeaderNames.ACCEPT -> MimeTypes.JSON)
```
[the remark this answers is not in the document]
drnaylor-oc: Accept is (well, it's good practice). Content-type is the one that isn't required.

app/v2/models/responses/hateoas/HateoasMovementResponse.scala
drnaylor-oc: I'd prefer to see this as pattern match cases -- I very much doubt we'll get another `MovementType`, but if we do (transit?), then we'll get warnings.

That said, can we just provide the movement type to the uri methods and get the message codes out from there? I don't recall if `MovementType` in the API includes `urlFragment` (that might just be in other services), but we can just use that instead of if statements like this (i.e. `s"/customs/transits/movements/${movementType.urlFragment}/...`)

test/routing/ArrivalsRouterSpec.scala
drnaylor-oc: It doesn't really matter here -- we just need to put _something_ into the URI of the request. We're not performing a full blown integration test so it doesn't route on this.

sonisn1988: if we are not using $endpoint.method then basically it is pointless to create it ?

drnaylor-oc: Oh, you're talking about the `method`. I thought you're talking about the URI.

test/routing/ArrivalsRouterSpec.scala
```diff
@@ -48,49 +48,56 @@ class ArrivalsRouterSpec extends AnyFreeSpec with Matchers with OptionValues wit
...
+          val result = call(sut.createArrivalNotification(), request)
```
drnaylor-oc, line 69: I see you're trying to check that all methods lead to the same place (as per the `for` loop above, but you're then only calling this one method on the actual router, and this is the key thing that needs to change.

test/v2/controllers/V2ArrivalsControllerSpec.scala
```diff
@@ -758,4 +758,71 @@ class V2ArrivalsControllerSpec
...
+        )
+
+      val request = FakeRequest(
+        GET,
+        routing.routes.ArrivalsRouter.getArrival(movementResponse._id.value).url,
+        headers = FakeHeaders(Seq(HeaderNames.ACCEPT -> VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON)),
+        AnyContentAsEmpty
+      )
+      val result = sut.getArrival(movementResponse._id)(request)
+      9
```
drnaylor-oc, line 777: I suspect you don't need this here!

app/v2/models/responses/hateoas/HateoasMovementUpdateResponse.scala
```diff
@@ -20,14 +20,15 @@ import play.api.libs.json.JsObject
...
+object HateoasMovementUpdateResponse extends HateoasResponse {
 
-  def apply(departureId: MovementId, messageId: MessageId): JsObject =
+  def apply(departureId: MovementId, messageId: MessageId, movementType: MovementType): JsObject =
     Json.obj(
       "_links" -> Json.obj(
-        "self"      -> Json.obj("href" -> messageUri(departureId, messageId)),
-        "departure" -> Json.obj("href" -> departureUri(departureId))
+        "self"                          -> Json.obj("href" -> getMessageUri(departureId, messageId, movementType)),
+        s"${movementType.movementType}" -> Json.obj("href" -> getMovementUri(departureId, movementType))
```
drnaylor-oc, line 31: This can just be `movementType.movementType` -- no need to wrap it in an s-interpolated string if you're just using that string.

test/routing/ArrivalsRouterSpec.scala
liandra-giacomine: Good spot, thanks Dan

test/v2/controllers/V2ArrivalsControllerSpec.scala
liandra-giacomine: Oh nooo hahaha thanks!

test/routing/ArrivalsRouterSpec.scala
```diff
@@ -48,75 +50,84 @@ class ArrivalsRouterSpec extends AnyFreeSpec with Matchers with OptionValues wit
...
-          }
+    .sample
+    .get
+
+  for (
+    callAndSutAndStatus <- Seq(
+      (routes.ArrivalsRouter.createArrivalNotification(), sut.createArrivalNotification(), ACCEPTED),
+      (routes.ArrivalsRouter.getArrival(id), sut.getArrival(id), OK),
+      (routes.ArrivalsRouter.getArrivalsForEori(), sut.getArrivalsForEori(), OK)
+    )
```
drnaylor-oc, line 66: I'm torn about this.

It's a clever way to reduce code (and I am indeed the sort of person who likes this sort of thing!), but it's not obvious what it's doing and why. Someone looking to add another endpoint that does version checks like this would probably get confused and/or write their own test that does exactly this because they don't understand it -- especially as it's outside of any test description block.

I think, for readability's sake, do something like this:

```scala
"routing to version 2 with a version 2 accept header" - {
    // second parameter is call by name so it only evaluates when we want it to
    def executeTest(callValue: (type), sutValue: => Future[Result], expectedStatus: Int) = {
        // test
    }

   "when creating an arrival notification" in executeTest(...)
   "when getting an arrival" in executeTest(...)
   "when getting arrivals for a given enrolment EORI" in executeTest(...)
}
```

test/routing/ArrivalsRouterSpec.scala
```diff
@@ -48,75 +50,84 @@ class ArrivalsRouterSpec extends AnyFreeSpec with Matchers with OptionValues wit
...
-        val result  = call(sut.getArrivalsForEori(), request)
-
-        status(result) mustBe OK
-        contentAsJson(result) mustBe Json.obj("version" -> 1) // ensure we get the unique value to verify we called the fake action
+  for (
+    callAndSutAndStatus <- Seq(
+      (routes.ArrivalsRouter.createArrivalNotification(), sut.createArrivalNotification(), ACCEPTED),
+      (routes.ArrivalsRouter.getArrival("123"), sut.getArrival("123"), OK),
+      (routes.ArrivalsRouter.getArrivalsForEori(), sut.getArrivalsForEori(), OK)
+    )
```
drnaylor-oc, line 98: Same here.

test/routing/ArrivalsRouterSpec.scala
liandra-giacomine: That's a good point Dan, thanks for the suggestion

drnaylor-oc (approved):

PR #333, opened 2022-11-08
Get arrival messages by arrival id (common-transit-convention-traders)

sonisn1988:


app/v2/connectors/PersistenceConnector.scala
```diff
@@ -241,6 +246,29 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
...
+  ): Future[Seq[MessageSummary]] = {
+    val url =
+      withReceivedSinceParameter(
+        appConfig.movementsUrl.withPath(movementsGetArrivalMessageIds(eori, movementId)).toUrl,
+        receivedSince
+      )
+
+    httpClientV2
+      .get(url"$url")
+      .addHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.XML)
```
liandra-giacomine, line 261: Get request so Accept header -> JSON?

app/v2/models/responses/hateoas/HateoasResponse.scala
```diff
@@ -19,15 +19,18 @@ package v2.models.responses.hateoas
...
 import v2.models.responses.hateoas.HateoasResponse.prefix
-
 import java.time.OffsetDateTime
 
 object HateoasResponse {
 
   lazy val prefix =
-    if (routing.routes.DeparturesRouter.submitDeclaration().url.startsWith(Constants.Context)) ""
+    if (
+      routing.routes.DeparturesRouter
```
liandra-giacomine, line 28: my lack of understanding here but why do we need this bit of code? I thought that if it started with context it would be calling .urlWithContext rather than .url

liandra-giacomine: Also this is a val...and you have implemented an OR here, so if one of them starts with context and the other does not, it will be duplicated for the one that already contains it

app/v2/models/responses/hateoas/HateoasResponse.scala
```diff
@@ -19,15 +19,18 @@ package v2.models.responses.hateoas
...
+        .submitDeclaration()
+        .url
+        .startsWith(Constants.Context) || routing.routes.ArrivalsRouter.createArrivalNotification().url.startsWith(Constants.Context)
+    ) ""
```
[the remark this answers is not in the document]
sonisn1988: removed

resources/public/api/conf/2.0/application.raml
```diff
@@ -448,4 +448,68 @@ traits:
...
+                                    "self":{
+                                       "href":"/customs/transits/movements/arrivals/123/messages/111"
+                                    },
+                                    "arrival":{
+                                       "href":"/customs/transits/movements/arrivals/123"
+                                    }
+                                },
+                                "id":"111",
+                                "arrivalId":"123",
+                                "received":"20220101 100903",
```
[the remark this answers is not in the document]
sonisn1988: updated

app/v2/connectors/PersistenceConnector.scala
sonisn1988: updated

app/v2/models/responses/hateoas/HateoasResponse.scala
sonisn1988: removed or condition as suggested by Dan

liandra-giacomine: @drnaylor-oc yeah makes sense - In my PR I am already making that modification btw @sonisn1988

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -63,11 +68,13 @@ trait ErrorTranslator {
...
         PresentationError.notFoundError(s"Departure movement with ID ${departureId.value} was not found")
-      case PersistenceError.MessageNotFound(movement, message) =>
+      case ArrivalNotFound(arrivalId) =>
+        PresentationError.notFoundError(s"Arrival movement with ID ${arrivalId.value} was not found")
+      case MessageNotFound(movement, message) =>
         PresentationError.notFoundError(s"Message with ID ${message.value} for movement ${movement.value} was not found")
-      case PersistenceError.DeparturesNotFound(eori) =>
+      case DeparturesNotFound(eori) =>
         PresentationError.notFoundError(s"Departure movement IDs for ${eori.value} were not found")
       case PersistenceError.ArrivalsNotFound(eori) =>
```
drnaylor-oc, line 79: if you're going to remove the `PersistenceError` qualifier, you should remove it from all the messages.

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -921,6 +921,176 @@ class PersistenceConnectorSpec
...
+        val result = persistenceConnector.getArrivalMessageIds(eori, arrivalId, None)
+        whenReady(result) {
+          _ mustBe messageResponse
+        }
+    }
+
+    "on successful return of arrival message IDs when filtering by received date is applied, return a success" in forAll(
+      arbitrary[EORINumber],
+      arbitrary[MovementId],
+      OffsetDateTime.now(ZoneOffset.UTC),
```
drnaylor-oc, line 961: You're probably better off using an arbitrary `OffsetDateTime` instead rather than `now`.

test/routing/ArrivalsRouterSpec.scala
```diff
@@ -24,6 +24,9 @@ import org.scalatest.matchers.must.Matchers
 import org.scalatestplus.mockito.MockitoSugar
 import play.api.http.HeaderNames
 import play.api.http.Status.ACCEPTED
+
+import play.api.http.Status.BAD_REQUEST
+
```
drnaylor-oc, line 29: Very much a nitpick, but can we make sure that imports like this don't have newlines between them for non-Java/Scala imports?

app/v2/controllers/ErrorTranslator.scala
sonisn1988: done

test/routing/ArrivalsRouterSpec.scala
sonisn1988: done

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -63,15 +70,17 @@ trait ErrorTranslator {
   implicit val persistenceErrorConverter = new Converter[PersistenceError] {
```
drnaylor-oc, line 71: This failed to compile because you've now got ambiguous imports. You should locally import the persistence errors here:

`import v2.models.errors.PersistenceError._`

and remove the imports in the global scope.

app/v2/controllers/ErrorTranslator.scala
```diff
@@ -80,7 +70,7 @@ trait ErrorTranslator {
         PresentationError.notFoundError(s"Departure movement IDs for ${eori.value} were not found")
       case ArrivalsNotFound(eori) =>
         PresentationError.notFoundError(s"Arrival movement IDs for ${eori.value} were not found")
-      case err: UnexpectedError => PresentationError.internalServiceError(cause = err.thr)
+      case err: errors.PersistenceError.UnexpectedError => PresentationError.internalServiceError(cause = err.thr)
```
drnaylor-oc, line 73: I'm not really happy with this, I'd much rather you went for the locally scoped imports like we've done in the other converters, above.

it/v2/connectors/PersistenceConnectorSpec.scala
sonisn1988: when I am using arbitrary OffsetDateTime it is failing with **request was not matched** and I can see some encoded value for time. Not sure how to use arbitrary here

drnaylor-oc: It makes sense.

```
GET                                                        | GET
/transit-movements/traders//movements/arrivals//messages/  | /transit-movements/traders//movements/arrivals//messages/<<<<< URL does not match
?receivedSince=+60085274-09-20T10:23:43.300069439-16:10:2  | ?receivedSince=%2B60085274-09-20T10:23:43.300069439-16:10
2                                                          | :22
```

`+` is becoming `%2B` when the request is made, which is entirely legitimate because URLs _should_ be URL encoded. However, and this is a problem with Wiremock I guess, or quirk at least, you're not looking for the URL encoded URL.

The problem with the time you gave is that you weren't potentially testing the + sign against the URLs. Now, thankfully, the code is doing the right thing in URL encoding the URL (see https://developer.service.hmrc.gov.uk/api-documentation/docs/api/service/common-transit-convention-traders/2.0#percentencoding-of-parameters-in-request-urls) -- it is your test that is listening on the wrong URL.

Fix how `targetUrlWithTime` is encoding the URL (do something like `${URLEncoder.encode(DateTimeFormatter.ISO_OFFSET_DATE_TIME.format(receivedSince), StandardCharsets.UTF_8.name())}`).

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -921,6 +922,178 @@ class PersistenceConnectorSpec
...
+      Gen.nonEmptyListOf(arbitrary[MessageSummary])
+    ) {
+      (eori, arrivalId, time, messageResponse) =>
+        server.stubFor(
+          get(
+            urlEqualTo(targetUrlWithTime(eori, arrivalId, time))
+          )
```
drnaylor-oc, line 971: I just came to the realisation (while washing up, of all things) what you're doing wrong here. You've used `urlEqualTo`, but in fact, you want `urlPathEqualTo` in combination with `withQueryParam`

```suggestion
        server.stubFor(
          get(
            urlPathEqualTo(targetUrl(eori, arrivalId))
          )
            .withQueryParam("receivedSince", equalTo(DateTimeFormatter.ISO_DATE_TIME.format(time)))
```

Note that this is an internal call where we explicity make a call in UTC. I'd forgotten that little bit when talking through this, BUT your test was wrong anyway because you were formatting using the `ISO_OFFSET_DATE_TIME`, not `ISO_DATE_TIME` which doesn't include the offset (because we don't need it).

Do this, delete `targetUrlWithTime`, we should be good. 

Before you push, compile your code and run the tests on it (use `./run_all_tests.sh`) to make sure this is all okay (just want to make sure nothing else comes out of the woodwork) -- and it might be worth testing your code with the regression tests too.

drnaylor-oc (approved):

PR #336, opened 2022-11-10
Added Get a cached arrival by arrival id to API definition 

liandra-giacomine:
I noticed that the departure ID examples in V2 was in the old format - int vs hex, so I updated those. I also fixed the response in one of the requests which was not reflecting what we actually return.


sonisn1988 (approved):
Perfect. Thank you :)

drnaylor-oc (approved):

PR #338, opened 2022-11-11
[CTCP-1648] added api for Get single arrival message by id

prashanthmamidi:


app/v2/controllers/V2ArrivalsController.scala
```diff
@@ -190,4 +196,17 @@ class V2ArrivalsControllerImpl @Inject() (
           )
     }
 
+  def getArrivalMessage(arrivalId: MovementId, messageId: MessageId): Action[AnyContent] =
+    (authActionNewEnrolmentOnly andThen acceptHeaderActionProvider()).async {
+      implicit request =>
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+        (for {
+          messageSummary          <- arrivalsService.getArrivalMessage(request.eoriNumber, arrivalId, messageId).asPresentation
+          formattedMessageSummary <- responseFormatterService.formatMessageSummary(messageSummary, request.headers.get(HeaderNames.ACCEPT).get)
```
sonisn1988, line 205: This Formatter is created for Json wrapped xml and that is not asked for this user story. if you are using this formatter then in Arrivals routers you should use case pattern matching Some(VERSION_2_ACCEPT_HEADER_PATTERN()) =>. 
@drnaylor-oc  is it ok to implement formatted Message summary as part of this?

drnaylor-oc: Note that it won't work for non 007 arrival messages at the moment anyway, because they need to be added to the converter!

...which I need to make a ticket for.

For what it's worth though -- I don't think there is any harm in leaving it if it's done. No point wasting work if it's been done.

liandra-giacomine: The issue with calling formatMessageSummary now is that it converts the xml to json if the accept header is set to +json, which is currently is. Once we implement CTCP-1705, we will introduce a breaking change where json+xml will return the current implementation (message in xml), and +json will be modified to return the message in json.

liandra-giacomine: @prashanthmamidi In that case, as mentioned by Sandeep, you need to update the router to route to V2 based on the V2 regex pattern rather than solely by +json header.

resources/public/api/conf/2.0/application.raml
```diff
@@ -616,4 +616,77 @@ traits:
                         description: The specified arrival does not exist, has been archived or is not available to the EORI number.
                         value:
                           code: NOT_FOUND
+        /{messageId}:
+          get:
+            displayName: Get a cached message related to a arrival and message ID
+            description: |
+              Retrieve all messages related to a specific arrival and message ID that have been received within the past 31 days.
+
+              For example, this could include ‘Declaration rejected’ or Movement Reference Number (MRN) allocated messages.
```
liandra-giacomine, line 625: This relates to departures. For arrivals, the P4 equivalent is:

"For example, this could include messages about route diversions or unloading permissions."

resources/public/api/conf/2.0/application.raml
```diff
@@ -616,4 +616,77 @@ traits:
...
+
+              Movements older than 31 days and their related messages are archived and cannot be returned by the API.
 
+               The response can be returned in one of the following formats based on the provided `Accept` header:
+
+               * `application/vnd.hmrc.2.0+json-xml`: JSON-wrapped XML, where the message is provided as an XML string sent as part of JSON payload containing other metadata.
+               * `application/vnd.hmrc.2.0+json`: JSON, where the message is provided as a sub-document sent as part of a JSON payload containing other metadata.
```
liandra-giacomine, line 632: Remove this bit until we implement the json conversion for arrivals

resources/public/api/conf/2.0/application.raml
```diff
@@ -616,4 +616,77 @@ traits:
...
+                            "href": "/customs/transits/movements/arrivals/62f4ebbbf581d4aa"
+                          }
+                        },
+                        "id": "62f4ebbb765ba8c2",
+                        "arrivalId": "62f4ebbbf581d4aa",
+                        "received": "2022-08-11T11:44:59.83705",
+                        "type": "IE007",
+                        "body": "<ncts:IE007 PhaseID=\"NCTS5.0\" xmlns:ncts=\"http://ncts.dgtaxud.ec\">...</ncts:IE007>"
+                      }
+                  application/json:
```
liandra-giacomine, line 661: Also remove this section for now

resources/public/api/conf/2.0/application.raml
prashanthmamidi: updated

resources/public/api/conf/2.0/application.raml
prashanthmamidi: updated

resources/public/api/conf/2.0/application.raml
prashanthmamidi: removed this

app/routing/ArrivalsRouter.scala
```diff
@@ -80,6 +85,26 @@ class ArrivalsRouter @Inject() (
       )
   }
 
+  def getArrivalMessage(arrivalId: String, messageId: String): Action[Source[ByteString, _]] = route {
+    case Some(VERSION_2_ACCEPT_HEADER_PATTERN()) =>
```
sonisn1988, line 89: As you have removed formatter from controller so you need to  use case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON) => and remove the VERSION_2_ACCEPT_HEADER_PATTERN
@liandra-giacomine / @drnaylor-oc I hope this is what we have agreed?

drnaylor-oc: No, this is correct. We want developers to test using `application/vnd.hmrc.2.0+json-xml` now, because that won't change.

resources/public/api/conf/2.0/application.raml
```diff
@@ -616,4 +616,56 @@ traits:
...
+                        "id": "62f4ebbb765ba8c2",
+                        "arrivalId": "62f4ebbbf581d4aa",
+                        "received": "2022-08-11T11:44:59.83705",
+                        "type": "IE007",
+                        "body":{
+                          "n1.CC007C": {
+                            "@PhaseID": "NCTS5.0",
+                            ...
+                          }
+                        }
```
drnaylor-oc, line 661: This bit isn't returning Json, it's returning XML. It should be something like:

```
body: "<ncts:CC007C PhaseID=\"NCTS5.0\" xmlns:ncts=\"http://ncts.dgtaxud.ec\">...</ncts:CC007C>"
```

app/routing/ArrivalsRouter.scala
sonisn1988: but this is going to work for application/vnd.hmrc.2.0+json as well as application/vnd.hmrc.2.0+json+xml if I am not mistaken? so should not better to use VERSION_2_ACCEPT_HEADER_VALUE_JSON and remove the action provider from controller?

drnaylor-oc: Correct, it will, and this is intended. We need it to work for the `json-xml` version more so than the `json` variant anyway, what you're suggesting will mean it _won't_ work for that.

resources/public/api/conf/2.0/application.raml
liandra-giacomine: @drnaylor-oc I think the XML body is already present under the json wrapped xml header application/json+xml. As per my comment above I think this should be removed until we implement the json header

liandra-giacomine: Oh I see what happened... @prashanthmamidi you removed the wrong section - should have removed the json rather than json wrapped xml https://github.com/hmrc/common-transit-convention-traders/pull/338/commits/5750b54a8b17f8f07cd673d626715f1e33af9654

drnaylor-oc: It's not, though I don't know if that was because it was removed before I looked at it.

The problem here is that `application/json+xml` isn't actually techincally a MIME type that is returned, `application/json` is -- what we have here is basically Json where we say XML is embedded as a string, so in the context of the response, this is correct. It's what I did on the equivalent endpoint for returning bot the Json and XML+Json endpoints.

It's not great, but it is what it is.

liandra-giacomine: @prashanthmamidi and this should have stayed there:
* `application/vnd.hmrc.2.0+json-xml`: JSON-wrapped XML, where the message is provided as an XML string sent as part of JSON payload containing other metadata.

drnaylor-oc: (looks like we responded at the same time again...)

liandra-giacomine: @drnaylor-oc I see what you're saying. For departures, it does look like application/json+xml and I wonder if it's because we can't duplicate application/json?

 https://github.com/hmrc/common-transit-convention-traders/blob/main/resources/public/api/conf/2.0/application.raml#L364

prashanthmamidi: updated the raml

sonisn1988 (dismissed):

resources/public/api/conf/2.0/application.raml
```diff
@@ -616,4 +616,53 @@ traits:
...
+               **Note:** You can view a transit movement only if your Economic Operators Registration and Identification (EORI) number matches the EORI number that created the transit declaration for that movement. Your EORI number is part of the bearer token that you use for API authentication.
+            is:
+              - acceptSingleMessageGetHeader
+              - acceptHeaderInvalid
+            (annotations.scope): "common-transit-convention-traders"
+            securedBy: [ sec.oauth_2_0: { scopes: [ "common-transit-convention-traders" ] } ]
+            responses:
+              200:
+                body:
+                  application/json+xml:
```
drnaylor-oc, line 642: Just going to check this works using the PREVIEW_RAML service manager profile.

resources/public/api/conf/2.0/application.raml
```diff
@@ -616,4 +616,53 @@ traits:
                         description: The specified arrival does not exist, has been archived or is not available to the EORI number.
                         value:
                           code: NOT_FOUND
+        /{messageId}:
```
drnaylor-oc, line 619: It's failed here:

```json
{
    "statusCode":500,
    "message":"Underlying error while parsing YAML syntax: 'while parsing a block mapping in 'reader', line 519, column 7:  get: expected <block end>, but found BlockMappingStart in 'reader', line 619, column 9: /{messageId}:' --  [line=619, col=1]"
}
```

You can check to see how changes work locally by:

* Running this project via `sbt run`
* Run the `PREVIEW_RAML` service manager profile (ignore the ASSETS_FRONTEND failure)
* Go to http://localhost:9680/api-documentation/docs/preview?url=http%3A%2F%2Flocalhost%3A9487%2Fapi%2Fconf%2F2.0%2Fapplication.raml

If it renders, it's good, if not, you'll get an error. There is probably an incorrect indent somewhere...

resources/public/api/conf/2.0/application.raml
drnaylor-oc: It works... and I'm a fool because I'm the one who did this in departures (and completely forgot). Sorry!

drnaylor-oc: Just to clarify why I did this in departures, it's because I wanted two different responses to appear on the definition page, and this was the only way I could do it. So it's fine -- but I wonder how this'll translate into OAS.

drnaylor-oc (approved):

PR #344, opened 2022-11-17
Allow message to be returned in JSON for GET single arrival message e…

sonisn1988:
…ndpoint

### DIFF
--- a/resources/public/api/conf/2.0/application.raml
+++ b/resources/public/api/conf/2.0/application.raml
@@ -394,9 +394,10 @@ traits:
                         "departure":{
                           "id": "/customs/transits/movements/departures/62f4ebbbf581d4aa"
                         },
-                        "id": "/customs/transits/movements/departures/62f4ebbbf581d4aa/messages/62f4ebbb765ba8c2",
+                        "id": "62f4ebbb765ba8c2",
+                        "departureId": "62f4ebbbf581d4aa",
                         "received": "2022-08-11T11:44:59.83705",
-                        "messageType": "IE015",
+                        "type": "IE015",
                         "body":{
                           "n1.CC015C": {
                             "@PhaseID": "NCTS5.0",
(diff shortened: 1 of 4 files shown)


test/v2/controllers/V2ArrivalsControllerSpec.scala
```diff
@@ -906,12 +907,15 @@ class V2ArrivalsControllerSpec
   }
 
   "GET /movements/arrivals/:arrivalId/messages/:messageId " - {
```
drnaylor-oc, line 909: I'd like tests for if the converter fails as well, please.

resources/public/api/conf/2.0/application.raml
```diff
@@ -391,12 +391,10 @@ traits:
...
                         },
-                        "departure":{
-                          "id": "/customs/transits/movements/departures/62f4ebbbf581d4aa"
-                        },
-                        "id": "/customs/transits/movements/departures/62f4ebbbf581d4aa/messages/62f4ebbb765ba8c2",
+                        "id": "62f4ebbb765ba8c2",
+                        "departureId": "62f4ebbbf581d4aa",
                         "received": "2022-08-11T11:44:59.83705",
-                        "messageType": "IE015",
+                        "type": "IE015",
```
drnaylor-oc, line 397: No, this should be `messageType`, as per https://github.com/hmrc/common-transit-convention-traders/blob/3991baefbe9aa83be97973a9cdc6c58335a6e74e/app/models/response/HateoasDepartureResponseMessage.scala#L40

resources/public/api/conf/2.0/application.raml
```diff
@@ -655,7 +653,30 @@ traits:
...
+                             "href": "/customs/transits/movements/arrivals/62f4ebbbf581d4aa/messages/62f4ebbb765ba8c2"
+                           },
+                           "arrival":{
+                             "href": "/customs/transits/movements/arrivals/62f4ebbbf581d4aa"
+                           }
+                         },
+                         "id": "62f4ebbb765ba8c2",
+                         "arrivalId": "62f4ebbbf581d4aa",
+                         "received": "2022-08-11T11:44:59.83705",
+                         "type": "IE007",
```
drnaylor-oc, line 673: Same here: https://github.com/hmrc/common-transit-convention-traders/blob/3991baefbe9aa83be97973a9cdc6c58335a6e74e/app/models/response/HateoasArrivalResponseMessage.scala#L40

sonisn1988: Liandra has modified it and it is now using HateoasMovementMessageResponse and there it is using  "type"                            -> messageSummary.messageType
.

resources/public/api/conf/2.0/application.raml
sonisn1988: Liandra has modified it and it is now using HateoasMovementMessageResponse and there it is using  "type"                            -> messageSummary.messageType

drnaylor-oc: Yup, you're right, my mistake, it is `type`. Carry on!

resources/public/api/conf/2.0/application.raml
drnaylor-oc: Yup!

liandra-giacomine (approved):

PR #347, opened 2022-12-05
added support to amend arrival

sonisn1988:


drnaylor-oc:
I'll try to look at this later, but your commit message(s) have caused some confusion with another ticket. This is more around adding new messages to arrivals -- not really amending the arrival itself (I guess it's technically is amending the arrival, but only as far as appending data, not changing exsiting data).

Can I ask that we include the ticket number in the PR summary/commit messages too, just so that Jira can pick them up?

resources/public/api/conf/2.0/application.raml
```diff
@@ -614,6 +614,56 @@ traits:
...
+                         code: NOT_FOUND
+             400:
+               body:
+                 application/json:
+                   type: types.errorResponse
+                   examples:
+                     badRequest:
+                       description: The arrival message type is not available within XML or the message failed schema validation.
+                       value:
+                         code: BAD_REQUEST
```
drnaylor-oc, line 666: @fmcgrath Would you be able to cast your eye over this to make sure it's okay for appending a message to an existing arrival? I'd imagine it should be mostly the same as v1 but I'd appreciate a look here, thanks!

fmcgrath: @drnaylor-oc - Looks OK. The only obvious differences between this version and v1 is the inclusion of `400` and `404` errors.

app/v2/models/request/MessageType.scala
```diff
@@ -95,12 +95,11 @@ object MessageType {
   )
 
   val messageTypesSentByArrivalTrader: Seq[MessageType] = Seq(
-    ArrivalNotification,
-    UnloadingRemarks,
-    InformationAboutNonArrivedMovement
+    UnloadingRemarks
```
liandra-giacomine, line 98: An ArrivalNotification is sent by the trader. It's the first message they send to start an arrival movement

liandra-giacomine (changes requested):

app/v2/models/request/MessageType.scala
sonisn1988: I have moved this to messageTypesSentToArrivalTrader Seq see below. Same is done for departure DeclarationData. As with this endpoint we need to first validate messageType using XmlParsrs messageTypeExtractor and ArriavlNotification is not correct messageType for this endpoint so I have moved out to messageTypesSentToArrivalTrader Seq

app/v2/services/XmlParsers.scala
```diff
@@ -20,24 +20,30 @@ import akka.NotUsed
...
+  def messageTypeExtractor(movementType: MovementType): Flow[ParseEvent, Either[ExtractionError, MessageType], NotUsed] = Flow[ParseEvent]
     .filter {
       case _: StartElement =>
         true
       case _ => false
     }
     .take(1)
     .map {
-      case s: StartElement =>
+      case s: StartElement if movementType == MovementType.Departure =>
```
liandra-giacomine, line 37: Instead of checking the movementType for every element that comes through, can we set a val here and reduce the duplication. For example:

`val messageTypes = if (movementType == MovementType.Departure) MessageType.messageTypesSentByDepartureTrader else MessageType.messageTypesSentByArrivalTrader`

`case s: StartElement => messageTypes.find(_.rootNode == s.localName)
          .map(Right(_))
          .getOrElse(Left(ExtractionError.MessageTypeNotFound(s.localName))) `

app/v2/services/XmlParsers.scala
```diff
@@ -20,24 +20,30 @@ import akka.NotUsed
...
         MessageType.messageTypesSentByDepartureTrader
           .find(_.rootNode == s.localName)
           .map(Right(_))
           .getOrElse(Left(ExtractionError.MessageTypeNotFound(s.localName)))
+      case s: StartElement if movementType == MovementType.Arrival =>
+        MessageType.messageTypesSentByArrivalTrader
+          .find(_.rootNode == s.localName)
```
drnaylor-oc, line 44: A little nitpicky, but there's a bit of duplication here that I'd like to see removed, just for ease of maintenance. Something like:

```scala
sentTypesFor(MovementType)
   .find(find(_.rootNode == s.localName) // etc...
```

or perhaps creating a link to the sent types in the `MovementType` object itself.

app/v2/models/request/MessageType.scala
drnaylor-oc: Which is wrong -- because `ArrivalNotifiaction` is _not_ sent to the trader. It was in the right place.

I see `DeclarationData` is in `messageTypesSentToDepartureTrader`... that's wrong too. I don't know _why_ it's there, but that needs to be fixed. I get why you want to do this, but we need to do this right (for both arrivals and departures)

app/v2/services/XmlParsers.scala
liandra-giacomine: @drnaylor-oc got there before me!

app/v2/models/request/MessageType.scala
sonisn1988: done the changes for both arrival and departure

app/v2/services/XmlParsers.scala
sonisn1988: done the changes as suggested

app/v2/services/XmlMessageParsingService.scala
```diff
@@ -52,26 +53,26 @@ class XmlMessageParsingServiceImpl @Inject() (implicit materializer: Materialize
 
   val messageSink = Sink.head[Either[ExtractionError, MessageType]]
 
-  private val messageTypeSink: Sink[ByteString, Future[Either[ExtractionError, MessageType]]] = Sink.fromGraph(
+  private def messageFlow(movementType: MovementType): Sink[ByteString, Future[Either[ExtractionError, MessageType]]] = Sink.fromGraph(
```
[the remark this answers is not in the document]
sonisn1988: done

app/v2/services/XmlParsers.scala
sonisn1988: done as suggested

app/v2/services/JsonMessageParsingService.scala
```diff
@@ -59,7 +59,11 @@ class JsonMessageParsingServiceImpl @Inject() (implicit materializer: Materializ
         .map {
           mt =>
             val root = mt.replace("\"", "")
-            MessageType.messageTypesSentByDepartureTrader.find(_.rootNode == root) match {
+            MessageType.messageTypesSentByDepartureTrader
+              .filterNot(
+                messageType => messageType == MessageType.DeclarationData
+              )
```
drnaylor-oc, line 65: I'm not convinced this is the right place to do this. This service is concerned with getting the message type out, and should not care about whether a message type is on an allow list.

Additionally, how will this work with arrivals?

I think a better way of doing this is to supply a `Seq[MessageType]` to the function that contains the message types you're willing to accept. That way, this code is agnostic to arrival/departure types, moving that responsibility to the appropriate place (the controller probably).

drnaylor-oc: Additionally, we're using this filtering code more than once. I wonder if we're better off just splitting the `messageTypesSentByDepartureTrader` (and arrivals version) into two lists:

```scala
val updateMessageTypesSentByDepartureTrader: Seq[MessageType] = Seq(
    // departure message types
)

val messageTypesSentByDepartureTrader: Seq[MessageType] = DeclarationData :+ updateMessageTypesSentByDepartureTrader
```

that way, we're not using `filterNot` all the time.

sonisn1988: Additionally, how will this work with arrivals? **This should be handled as part of Json Arrival json implementation and Prashant is working on on that part.**

I think a better way of doing this is to supply a Seq[MessageType] to the function that contains the message types you're willing to accept. That way, this code is agnostic to arrival/departure types, moving that responsibility to the appropriate place (the controller probably).    **This I have already done for xml under XmlParsers**
Also I have changed the MessageType list as suggested above

app/v2/services/JsonMessageParsingService.scala
```diff
@@ -59,11 +59,7 @@ class JsonMessageParsingServiceImpl @Inject() (implicit materializer: Materializ
         .map {
           mt =>
             val root = mt.replace("\"", "")
-            MessageType.messageTypesSentByDepartureTrader
-              .filterNot(
-                messageType => messageType == MessageType.DeclarationData
-              )
-              .find(_.rootNode == root) match {
+            MessageType.updateMessageTypesSentByDepartureTrader.find(_.rootNode == root) match {
```
liandra-giacomine, line 62: @sonisn1988  It looks like we still have it as the responsibility of the json/xml parsers to determine whether the message type is valid. It seems to go against what @drnaylor-oc suggested as it checks that the message type fits the expected list rather than just identifying a message type (from a list of all message types?) and checking if it's the expected message type in the controller. I think Dan's suggestion would mean that we just return the value in the StartElement/Json root node and return that to the controller. The controller would then check if it fits the expected message types for the movement. Perhaps we could do this as a tech debt to prevent not completing this prior sprint end?

drnaylor-oc: Rather, it's not the responsibility of this service to determine which list of message types to check against. I'd argue that the Seq of messages it should be able to get should be supplied to this service, this shouldn't determine it itself.

I'm basically suggesting that `MessageType.updateMessageTypesSentByDepartureTrader` is really something that should be passed in as an argument to the function, i.e. the controller determines the valid values, and this service uses that list to find if that's the case.

However, we're going to end up going around in circles on this one, as this isn't the Json ticket but the XML one, let's just get this done here.

sonisn1988: Thanks @drnaylor-oc / @liandra-giacomine I did the changes for both xml and json as suggested above. Please see below with this commit
https://github.com/hmrc/common-transit-convention-traders/pull/347/commits/91d9905f58d45b651237de07a9a3b298d55d62dd

test/v2/services/XmlParsersSpec.scala
```diff
@@ -31,33 +31,81 @@ import scala.xml.NodeSeq
...
+        </TransitOperation>
+      </ncts:CC007C>
+
     val invalidMessageType: NodeSeq =
-      <HolderOfTheTransitProcedure>
-        <CC013C>IE013</CC013C>
-      </HolderOfTheTransitProcedure>
+      <TransitOperation>
+          <MRN>27WF9X1FQ9RCKN0TM3</MRN>
+        </TransitOperation>
```
drnaylor-oc, line 65: I'm curious, why did you change the invalid message type to this? What happens if you leave it as is? `HolderOfTheTransitProcedure` on the root is an invalid message type after all.

liandra-giacomine (dismissed):
LGTM

test/v2/services/XmlParsersSpec.scala
sonisn1988: Mistakenly copy pasted, As added few test cases for arrivals. Reverted back to original.

liandra-giacomine (approved):

PR #350, opened 2022-12-09
CTDA-1973

liandra-giacomine:
Set received value returning in GET messages request to be when we receive the message rather than when it was generated


sonisn1988 (dismissed):
LGTM

drnaylor-oc (dismissed):

drnaylor-oc (approved):

PR #352, opened 2022-12-15
Combine departure and arrival endpoints into movements endpoints 

liandra-giacomine:
Updated transit-movements get departure by eori endpoint and removed duplication in departures and arrivals implementation


app/v2/controllers/ErrorTranslator.scala
```diff
@@ -59,16 +59,12 @@ trait ErrorTranslator {
     import v2.models.errors.PersistenceError._
 
     def convert(persistenceError: PersistenceError): PresentationError = persistenceError match {
-      case DepartureNotFound(departureId) =>
-        PresentationError.notFoundError(s"Departure movement with ID ${departureId.value} was not found")
-      case ArrivalNotFound(arrivalId) =>
-        PresentationError.notFoundError(s"Arrival movement with ID ${arrivalId.value} was not found")
+      case MovementNotFound(departureId, movementType) =>
```
sonisn1988, line 62: Please use movementId

app/v2/controllers/V2MovementsController.scala
```diff
@@ -0,0 +1,328 @@
...
+    with ContentTypeRouting
+    with HasActionMetrics {
+
+  lazy val sCounter: Counter = counter(s"success-counter")
+  lazy val fCounter: Counter = counter(s"failure-counter")
+
+  def createMovement(movementType: MovementType): Action[Source[ByteString, _]] =
+    movementType match {
+      case MovementType.Arrival =>
+        contentTypeRoute {
```
sonisn1988, line 103: I think best to pass movementType as parameter with submitArrivalNotificationXML and submitArrivalNotificationJSON and use it later instead of hardcoding every time later in code

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -1444,7 +1443,7 @@ class PersistenceConnectorSpec
 
         implicit val hc = HeaderCarrier()
         val r = persistenceConnector
-          .getArrivalMessage(eori, arrivalId, messageId)
+          .getMessage(eori, MovementType.Arrival, arrivalId, messageId)
```
sonisn1988, line 1446: I am not sure if I am right here. Can we not combine arrival/departure using forEach with movementType list ? so that we can reduce test cases code as well

test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -0,0 +1,2126 @@
...
+
+    }
+
+    "must return UNSUPPORTED_MEDIA_TYPE when the content type is invalid" in {
+      val standardHeaders = FakeHeaders(
+        Seq(HeaderNames.ACCEPT -> "application/vnd.hmrc.2.0+json", HeaderNames.CONTENT_TYPE -> "invalid", HeaderNames.CONTENT_LENGTH -> "1000")
+      )
+
+      val json    = Json.stringify(Json.obj("CC015" -> Json.obj("SynIdeMES1" -> "UNOC")))
+      val request = fakeCreateMovementRequest("POST", standardHeaders, Source.single(json), MovementType.Arrival)
```
sonisn1988, line 1284: same with this class file as well. can we not combine arrival/departure with movemenType list

test/v2/services/MovementsServiceSpec.scala
```diff
@@ -0,0 +1,558 @@
...
+      Gen.listOfN(3, arbitrary[MovementSummary]),
+      Gen.option(arbitrary[OffsetDateTime]),
+      arbitrary[EORINumber]
+    ) {
+
+      (expected, updatedSinceMaybe, eori) =>
+        when(mockConnector.getMovements(eori, MovementType.Arrival, updatedSinceMaybe))
+          .thenReturn(Future.successful(expected))
+
+        val result = sut.getMovements(eori, MovementType.Arrival, updatedSinceMaybe)
```
sonisn1988, line 426: same with this file. Please try to combine arrival/departure test cases based on movementType. Ok with if not possible to combine. just wanted to highlight to reduce the lines of code.

app/v2/controllers/V2MovementsController.scala
liandra-giacomine: We need to determine what the movement type is at the router level, for the distinct endpoints

sonisn1988: yes agree. By passing it to submitArrivalNotificationXML and submitArrivalNotificationJSON we can achieve it.

liandra-giacomine: Oh I see what you're saying - but no, it's hard coded because it'll never change from being an arrival and it isn't hard coded from the router because it could be an arrival or departure.

sonisn1988: Fine. No problem

app/v2/controllers/V2MovementsController.scala
```diff
@@ -0,0 +1,328 @@
...
+  private def submitDepartureDeclarationJSON(): Action[Source[ByteString, _]] =
+    (authActionNewEnrolmentOnly andThen messageSizeAction()).stream {
+      implicit request =>
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+
+        (for {
+          _ <- validationService.validateJson(MessageType.DeclarationData, request.body).asPresentation
+          _ = auditService.audit(AuditType.DeclarationData, request.body, MimeTypes.JSON)
+          xmlSource         <- conversionService.jsonToXml(MessageType.DeclarationData, request.body).asPresentation
+          declarationResult <- persistAndSendToEIS(xmlSource, MovementType.Departure, MessageType.DeclarationData)
```
drnaylor-oc, line 142: I don't quite know why this has happened, but I thought the whole idea of `persistAndSendToEIS` was to unify the XML and Json routes -- but that's not happened here (it was the same when I was looking at the attach routes, it caught me out as I thought I only had to update something in one place).

Can we look to see if we can make the XML route also use this method, so that we don't have the implementation multiple times when we don't need to?

app/v2/controllers/V2MovementsController.scala
```diff
@@ -0,0 +1,328 @@
...
+    def handleXml(movementId: MovementId, eoriNumber: EORINumber, messageType: MessageType, src: Source[ByteString, _])(implicit
+      hc: HeaderCarrier
+    ): EitherT[Future, PresentationError, UpdateMovementResponse] =
+      withReusableSource(src) {
+        source =>
+          for {
+            _              <- validationService.validateXml(messageType, source).asPresentation(jsonToXmlValidationErrorConverter, materializerExecutionContext)
+            updateResponse <- movementsService.updateMovement(movementId, movementType, messageType, source).asPresentation
+            _ = pushNotificationsService.update(movementId)
+            _ <- routerService.send(messageType, eoriNumber, movementId, updateResponse.messageId, source).asPresentation
```
drnaylor-oc, line 286: I made a change in a previous PR to ensure that this logic (which is the same for XML and Json routes) was put in its own method -- it seems to have been removed here and we have duplication across both routes -- can we put the single method back?

drnaylor-oc (approved):

test/v2/controllers/V2MovementsControllerSpec.scala
liandra-giacomine: I am doing that for the GET requests, but since the implementation for creating a movement varies for departures and arrivals, I think it makes sense to keep their tests separate

PR #356, opened 2022-12-21
CTCP-1956

sonisn1988:
common transit convention traders - Add movement EORI filter to get all cached arrivals/departures endpoint

### DIFF
--- a/app/v2/connectors/PersistenceConnector.scala
+++ b/app/v2/connectors/PersistenceConnector.scala
@@ -25,6 +25,7 @@ import com.kenshoo.play.metrics.Metrics
 import config.AppConfig
 import config.Constants
 import io.lemonlabs.uri.Url
+import io.lemonlabs.uri.config.ExcludeNones
 import metrics.HasMetrics
 import metrics.MetricsKeys
 import play.api.Logging
@@ -181,7 +182,7 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
     hc: HeaderCarrier,
     ec: ExecutionContext
   ): Future[Seq[MovementSummary]] = {
-    val urlWithOptions = withOptionalParameter(
+    val urlWithOptions = withParameters(
       appConfig.movementsUrl.withPath(getAllMovementsUrl(eori, movementType)),
       updatedSince,
       movementEORI
@@ -227,13 +228,15 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
       )
       .getOrElse(urlPath)
 
-  private def withOptionalParameter(urlPath: Url, updatedSince: Option[OffsetDateTime], movementEORI: Option[EORINumber]) =
-    (updatedSince, movementEORI) match {
-      case (Some(updatedSince), Some(movementEORI)) =>
-        urlPath.addParams("updatedSince" -> DateTimeFormatter.ISO_DATE_TIME.format(updatedSince), "movementEORI" -> movementEORI.value)
-      case (Some(updatedSince), None) => urlPath.addParam("updatedSince" -> DateTimeFormatter.ISO_DATE_TIME.format(updatedSince))
-      case (None, Some(movementEORI)) => urlPath.addParam("movementEORI" -> movementEORI.value)
-      case (None, None)               => urlPath
-    }
+  def withParameters(urlPath: Url, updatedSince: Option[OffsetDateTime], movementEORI: Option[EORINumber]) =
+    urlPath
+      .withConfig(urlPath.config.copy(renderQuery = ExcludeNones))
+      .addParam(
+        "updatedSince",
+        updatedSince.map(
+          time => DateTimeFormatter.ISO_DATE_TIME.format(time)
+        )
+      )
+      .addParam("movementEORI", movementEORI.map(_.value))
 
 }
--- a/app/v2/connectors/PersistenceConnector.scala
+++ b/app/v2/connectors/PersistenceConnector.scala
@@ -25,8 +25,6 @@ import com.kenshoo.play.metrics.Metrics
 import config.AppConfig
 import config.Constants
 import io.lemonlabs.uri.Url
-import io.lemonlabs.uri.config.ExcludeNones
-import io.lemonlabs.uri.config.UriConfig
 import metrics.HasMetrics
 import metrics.MetricsKeys
 import play.api.Logging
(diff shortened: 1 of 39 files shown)


app/v2/connectors/PersistenceConnector.scala
```diff
@@ -177,18 +177,18 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
       }
   }
 
-  override def getMovements(eori: EORINumber, movementType: MovementType, updatedSince: Option[OffsetDateTime])(implicit
+  override def getMovements(eori: EORINumber, movementType: MovementType, updatedSince: Option[OffsetDateTime], movementEORI: Option[EORINumber])(implicit
     hc: HeaderCarrier,
     ec: ExecutionContext
   ): Future[Seq[MovementSummary]] = {
-    val url = withDateTimeParameter(
+    val urlWithDateOption = withDateTimeParameter(
```
liandra-giacomine, line 184: can we please move the logic in withDateTimeParameter and withMovementEORIParameter to the V2BaseConnector, so that we can call getAllMovementsUrl with the movementEori and updatedSince and get the path back?

Also, do we even need those separate methods or does addQueryParam consider Optional values

build.sbt
```diff
@@ -27,7 +27,8 @@ lazy val microservice = Project(appName, file("."))
       "models._",
       "models.domain._",
       "models.Binders._",
-      "java.time.OffsetDateTime"
+      "java.time.OffsetDateTime",
+      "v2.models._"
```
liandra-giacomine, line 31: If we only need the EORI, only import the EORI

liandra-giacomine (commented):
Looks good, just a couple of points

app/v2/connectors/PersistenceConnector.scala
sonisn1988: I don't think so that is better way to do this to move to V2BaseConnector. As with V2BaseConnector with all the api we are getting the base path(UrlPath). Same way it is done for others like for updatedSince and receivedSince optional param we have used those optional parameters in PersistenceConnector. Also if we are moving withDateTimeParameter and withMovementEORIParameter to V2BaseConnector it will little bit complicate the things.
So I would say better to keep it as.

Yes we need two separate method as in case of we are not passing movementEORI it should call base uri without movementEORI. But when I tried with **withQueryParam** and in case of not passing any value to movementEORI it is appending movementEORI with the request url and so not returning anything from backend api.

liandra-giacomine: I see what you mean about it not being a good idea to move it to the BaseConnector, but I still think we can simplify the code by using patten matching:

`   val url = appConfig.movementsUrl.withPath(getAllMovementsUrl(eori, movementType))

    val urlWithQueryParams = (updatedSince, movementEORI) match {
      case (Some(datetime), Some(eori)) => url.addParams("updatedSince" -> DateTimeFormatter.ISO_DATE_TIME.format(datetime), "movementEORI" -> eori.value)
      case (Some(datetime), None)       => url.addParams("updatedSince" -> DateTimeFormatter.ISO_DATE_TIME.format(datetime))
      case (None, Some(eori))           => url.addParams("movementEORI" -> eori.value)
      case (None, None)                 => url
    }`

sonisn1988: done as suggested. Thanks

app/v2/connectors/PersistenceConnector.scala
```diff
@@ -228,4 +227,13 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
...
       .getOrElse(urlPath)
 
+  private def withOptionalParameter(urlPath: Url, updatedSince: Option[OffsetDateTime], movementEORI: Option[EORINumber]) =
+    (updatedSince, movementEORI) match {
+      case (Some(updatedSince), Some(movementEORI)) =>
+        urlPath.addParams("updatedSince" -> DateTimeFormatter.ISO_DATE_TIME.format(updatedSince), "movementEORI" -> movementEORI.value)
+      case (Some(updatedSince), None) => urlPath.addParam("updatedSince" -> DateTimeFormatter.ISO_DATE_TIME.format(updatedSince))
+      case (None, Some(movementEORI)) => urlPath.addParam("movementEORI" -> movementEORI.value)
+      case (None, None)               => urlPath
+    }
```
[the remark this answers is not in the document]
sonisn1988: Thanks. Done the changes

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -680,7 +709,34 @@ class PersistenceConnectorSpec
...
+      Gen.listOfN(3, arbitraryMovementSummary.arbitrary),
+      None,
+      arbitrary[EORINumber]
+    ) {
+      (movementSummaryList, updatedSince, movementEORI) =>
+        server.stubFor(
+          get(
+            urlPathEqualTo(targetUrl(eori))
+          )
+            .withQueryParam("movementEORI", equalTo(movementEORI.value))
```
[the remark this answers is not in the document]
sonisn1988: done

it/v2/connectors/PersistenceConnectorSpec.scala
```diff
@@ -1224,7 +1312,34 @@ class PersistenceConnectorSpec
...
```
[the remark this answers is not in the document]
sonisn1988: done

resources/public/api/conf/2.0/application.raml
```diff
@@ -151,6 +153,11 @@ traits:
           type: datetime
           description: This is an optional query parameter that you can use to filter for departure declarations updated after a specific date and time.
           example: 2021-06-21T09:00+00:00
+        movementEORI:
+          required: false
+          type: string
+          description: This is an optional query parameter that you can use to filter for departure declarations that were submitted with a specified movement EORI.
+          example: ezv3Z
```
liandra-giacomine, line 160: "[A-Z]{2}[!-~]{1,15}" Looks like first two letters should be capital letters

drnaylor-oc: Yep, best for this to be a representative example, so something like `GB1234567890`

liandra-giacomine (approved):

liandra-giacomine (changes requested):
Need to understand why an empty value for movement eori is returning 404

drnaylor-oc:
> Need to understand why an empty value for movement eori is returning 404

It'll be represented by `Some("")` in the argument, so filtering on EORIs of "" -- of which there aren't any. Perhaps add a test for this -- though this would really be testing the framework.

liandra-giacomine (dismissed):
Working as expected, my bad

liandra-giacomine (approved):

PR #359, opened 2022-12-28
Ignore movementEORI query param if value is not passed in request URI

sonisn1988:


drnaylor-oc:
We're not going to do it this way after discussion -- closing as a result.

PR #366, opened 2023-01-03
CTCP-2138: Mention PPNS in the API defintion

drnaylor-oc:
Not for merging until we've gotten our PPNS integration up and running.

### DIFF
--- a/resources/public/api/conf/2.0/application.raml
+++ b/resources/public/api/conf/2.0/application.raml
@@ -298,7 +298,7 @@ traits:
     
             For example, if you want to cancel a ‘Declaration Data’ E_DEC_DAT (IE015) message, you must send a ‘Declaration Invalidation Request’ E_DEC_INV (IE014) message to the Office of Departure.
     
-            A successful response will contain a URI and a message ID related to the departure, which you can use to retrieve the message status.
+            A successful response will contain a URI and a message ID related to the departure that you can use to get the message status.
     
             **Note:** If an Economic Operators Registration and Identification (EORI) number has not been assigned to you, an HTTP 403 Forbidden error is returned.
           is:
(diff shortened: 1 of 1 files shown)


resources/public/api/conf/2.0/application.raml
```diff
@@ -102,7 +102,7 @@ traits:
 
         This request is also called an ‘Declaration Data’ E_DEC_DAT (IE015) message.
 
-        A successful response will contain a URI and a departure ID which you can use to retrieve messages related to the movement.
+        A successful response will contain a URI which you can use to retrieve the message status. A successful response can also include an optional box ID, which you can use to match this movement to any messages received from the Push Pull Notifications API.
```
fmcgrath, line 105: Suggestion:
A successful response will contain a URI that you can use to retrieve the message status.

The issue here is grammar around usage of that/which. "That" is used in a defining sense whereas "which" is usually used in a commenting sense. If "which" is a better choice in a sentence, it must usually be prefixed by a comma. In this case, I think that "that" is a better fit for the sentence.

resources/public/api/conf/2.0/application.raml
```diff
@@ -422,7 +423,7 @@ traits:
 
         This request is also called an ‘Arrival Notification’ E_ARR_NOT (IE007) message.
 
-        A successful response will contain a URI and an arrival ID which you can use to retrieve messages related to the movement.
+        A successful response will contain a URI which you can use to retrieve the message status. A successful response can also include an optional box ID, which you can use to match this movement to any messages received from the Push Pull Notifications API.
```
fmcgrath, line 427: Same issue as above.

Suggestion:
A successful response will contain a URI that you can use to get the message status.

fmcgrath (commented):
Two minor problems with a sentence that appears twice.

drnaylor-oc:
@fmcgrath Updated as per your suggestions, thanks! Mind doing another quick sweep?

fmcgrath (approved):
All good.

drnaylor-oc:
Thanks!

Will merge once we can deploy the push-notifications service to trader test, currently gremlins in the deploy system.

PR #370, opened 2023-01-23
[CTCP-2186] - changes to initiate call to upscan for large messages

prashanthmamidi:


test/v2/models/responses/hateoas/HateoasNewMovementResponseSpec.scala
```diff
@@ -71,4 +74,72 @@ class HateoasNewMovementResponseSpec extends AnyFreeSpec with Matchers with Scal
...
+            "key"                                 -> "b72d9aea-fdb9-40f1-800c-3612154baf07",
+            "acl"                                 -> "private",
+            "x-amz-signature"                     -> "xxxx",
+            "error_action_redirect"               -> "https://myservice.com/errorPage",
+            "x-amz-meta-session-id"               -> "3506d041-ba59-41ee-bb2c-bf0363163be3",
+            "x-amz-meta-consuming-service"        -> "PostmanRuntime/7.29.2",
+            "policy"                              -> "eyJjb25kaXRpb25zIjpbWyJjb250ZW50LWxlbmd0aC1yYW5nZSIsMCwxMDI0XV19"
+          )
+        )
+      )
```
drnaylor-oc, line 114: I'd like to see another test where the fields have changed to ensure that we're passing through whatever upscan-initiate gives us.

drnaylor-oc (commented):
Initial thoughts.

app/v2/connectors/PersistenceConnector.scala
```diff
@@ -117,6 +122,27 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
...
+          .execute[HttpResponse]
+          .flatMap {
+            response =>
+              response.status match {
+                case OK => response.as[MovementResponse]
+                case _ =>
+                  response.error
+              }
+          }
+    }
```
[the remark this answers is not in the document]
liandra-giacomine: We don't need to remove it since we already check for a None in contentTypeRouting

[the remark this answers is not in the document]
drnaylor-oc: Good point.

app/metrics/MetricsKeys.scala
```diff
@@ -54,6 +54,10 @@ object MetricsKeys {
     val Update = "push-notifications-backend-update"
   }
 
+  object UpscanInitiateBacked {
```
liandra-giacomine, line 57: UpscanInitiateBacked should be UpscanInitiateBackend, can you also please fix PushNotificationsBacked...should be PushNotificationsBackend

app/v2/connectors/PersistenceConnector.scala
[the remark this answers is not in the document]
prashanthmamidi: yes, Initially started with the same approach to have an Option for Source -  if it's for small message we need to have something like  ```
 httpClientV2
          .post(url"$url")
          .transform(_.addHttpHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.XML))
          .withBody(source)
          .execute[HttpResponse] 
```
and for large message,    ```
httpClientV2
          .post(url"$url")
          .execute[HttpResponse]
```
instead of conditionally handling this, thought of better to separate method - so we can easily manage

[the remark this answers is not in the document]
drnaylor-oc: although, that said, I looked into the implementation of HMRC's withBody and it does a bit more than just set the body, so actually, yeah, we'd need to separate it out. 

So, something like

```scala
  override def postMovement(eori: EORINumber, movementType: MovementType, source: Option[Source[ByteString, _]])(implicit
    hc: HeaderCarrier,
    ec: ExecutionContext
  ): Future[MovementResponse] =
    withMetricsTimerAsync(MetricsKeys.ValidatorBackend.Post) {
      _ =>
        val url = appConfig.movementsUrl.withPath(postMovementUrl(eori, movementType))

        val httpClient = httpClientV2.post(url"$url")
        (source match {
          case Some(x) =>
            httpClient.transform(_.addHttpHeaders(HeaderNames.CONTENT_TYPE -> MimeTypes.XML)).withBody(x)
          case None => httpClient
        }).execute[HttpResponse]
          .flatMap {
            response =>
              response.status match {
                case OK => response.as[MovementResponse]
                case _ =>
                  response.error
              }
          }
    }
```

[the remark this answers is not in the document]
liandra-giacomine:
```
  .execute[HttpResponse]
          .flatMap {
            response =>
              response.status match {
                case OK => response.as[MovementResponse]
                case _ =>
                  response.error
              }
          }
```

Can probably just be .execute[MovementResponse]?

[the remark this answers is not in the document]
drnaylor-oc: Yes, it could be.

app/v2/services/UpscanService.scala
```diff
@@ -0,0 +1,65 @@
...
+  upscanConnector: UpscanConnector,
+  appConfig: AppConfig
+) extends UpscanService {
+
+  override def upscanInitiate()(implicit
+    headerCarrier: HeaderCarrier,
+    executionContext: ExecutionContext
+  ): EitherT[Future, UpscanInitiateError, UpscanInitiateResponse] = {
+
+    val upscanInitiate = UpscanInitiate("https://myservice.com/callback", None, None, None, None) // TODO callback url will be configured later
```
liandra-giacomine, line 53: @drnaylor-oc I think we should be setting the max file size that we accept?

drnaylor-oc: Actually, yes, that's a good idea. It should be configurable, but it should be set to 22MB for now (we may drop it to 20MB to give us headroom)

conf/application.conf
```diff
@@ -225,6 +225,12 @@ microservice {
       port = 9508
       enabled = true
     }
+
+    transit-movements-upscan-initiate {
```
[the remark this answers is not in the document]
derekspalmer: Or even just upscan ?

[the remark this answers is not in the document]
drnaylor-oc: No, the service we'll be connecting to is specifically called `upscan-initiate`. We should have this disambiguation in there in case we ever need to connect to another Upscan service, now or in the future.

conf/application.conf
```diff
@@ -225,6 +225,12 @@ microservice {
...
+      protocol = "http"
+      host = "localhost"
+      port = 9570
+    }
```
drnaylor-oc, line 233: We'll need to add this to the `app-config-base` repository as well with the correct remote values.

conf/application.conf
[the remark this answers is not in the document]
prashanthmamidi: ok, we rename to `upscan-initiate`

app/v2/connectors/PersistenceConnector.scala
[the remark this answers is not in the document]
prashanthmamidi: Updated the code to use Option[Source]

app/config/AppConfig.scala
```diff
@@ -38,7 +38,9 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: CTCServicesCon
   val routerUrl                = Url.parse(servicesConfig.baseUrl("transit-movements-router"))
   val auditingUrl              = Url.parse(servicesConfig.baseUrl("transit-movements-auditing"))
   val pushNotificationsUrl     = Url.parse(servicesConfig.baseUrl("transit-movements-push-notifications"))
-  val upscanInitiateUrl        = Url.parse(servicesConfig.baseUrl("transit-movements-upscan-initiate"))
+  val upscanInitiateUrl        = Url.parse(servicesConfig.baseUrl("upscan-initiate"))
+  val upscanCallbackUrl        = servicesConfig.config("upscan-initiate").get[String]("callbackUrl")
+  val upscanMaximumFileSize    = servicesConfig.config("upscan-initiate").get[Long]("maximumFileSize") * 1000000  // 1MB = 1000000 bytes
```
liandra-giacomine, line 43: should be 22 mb

test/v2/services/MovementsServiceSpec.scala
```diff
@@ -343,19 +368,44 @@ class MovementsServiceSpec
...
+        .thenReturn(Future.failed(upstreamErrorResponse))
+      val result                                               = sut.createMovement(EORINumber("1"), MovementType.Arrival, Some(invalidRequest))
+      val expected: Either[PersistenceError, MovementResponse] = Left(PersistenceError.UnexpectedError(Some(upstreamErrorResponse)))
+      whenReady(result.value) {
+        _ mustBe expected
+      }
+    }
+  }
+
+  "Create Arrival notification for Large Messages" - {
```
liandra-giacomine, line 391: we haven't implemented this yet, so how is it working? Is there something wrong with how we're testing for a large message here?

liandra-giacomine: We're setting any[HeaderCarrier] but we need to ensure that we aren't sending a content-type header, would setting it to eqTo instead work?

conf/application.conf
```diff
@@ -225,6 +225,14 @@ microservice {
       port = 9508
       enabled = true
     }
+
+    upscan-initiate {
+      protocol = "http"
+      host = "localhost"
+      port = 9570
+      callbackUrl = "https://myservice.com/callback"
+      maximumFileSize = 22
```
drnaylor-oc, line 234: Can we add a comment in the config to make it clear that we're specifying this in megabytes, just so it's clear at one glance?

(I was kind of wondering if we should specify this in bytes or KB instead, megabytes is reasonable, as long as this accepts a decimal... may be better to be KB at least.)

app/config/AppConfig.scala
```diff
@@ -38,6 +38,9 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: CTCServicesCon
   val routerUrl                = Url.parse(servicesConfig.baseUrl("transit-movements-router"))
   val auditingUrl              = Url.parse(servicesConfig.baseUrl("transit-movements-auditing"))
   val pushNotificationsUrl     = Url.parse(servicesConfig.baseUrl("transit-movements-push-notifications"))
+  val upscanInitiateUrl        = Url.parse(servicesConfig.baseUrl("upscan-initiate"))
+  val upscanCallbackUrl        = servicesConfig.config("upscan-initiate").get[String]("callbackUrl")
+  val upscanMaximumFileSize    = servicesConfig.config("upscan-initiate").get[Long]("maximumFileSize") * 1000000  // 1MB = 1000000 bytes
```
drnaylor-oc, line 43: This should be a long overall, but the config option should be a double if we're going to specify it in megabytes, just in case we want fractional megabytes.

Alternatively, we could change the configuration to be in kilobytes, and then leaving this as long (and a whole number) is viable (and thus probably more accurate). Happy either way, really.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -124,7 +130,23 @@ class V2MovementsControllerImpl @Inject() (
           movementResponse <- persistAndSendToEIS(request.body, MovementType.Departure, MessageType.DeclarationData)
         } yield movementResponse).fold[Result](
           presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
-          response => Accepted(HateoasNewMovementResponse(response.movementId, response.boxResponse, MovementType.Departure))
+          response => Accepted(HateoasNewMovementResponse(response.movementId, response.boxResponse, None, MovementType.Departure))
+        )
+    }
+
+  private def submitDepartureDeclarationLargeXML(): Action[Source[ByteString, _]] =
+    (authActionNewEnrolmentOnly andThen messageSizeAction()).async(streamFromMemory) {
```
sonisn1988, line 138: @drnaylor-oc do we need to include andThen messageSizeAction()) ? Because we are not passing any payload here so we should not include it? if really want to include it then basically we need to customize that action and if payload is not empty in case of large message we should report error to end user?

drnaylor-oc: Yeah, that's a good point, `messageSizeAction` shouldn't be here at all.

prashanthmamidi: Don't think the code reaches here - If I removed contentLength header - I always get `501 Not Implemented` - anyways, will remove the action `messageSizeAction`

test/v2/services/MovementsServiceSpec.scala
prashanthmamidi: For transitmovements, it deals with based on movement - as the response for messageID is optional now - which impact both Departure and Arrival

conf/application.conf
prashanthmamidi: will add a comment to reflect it has 22 MB

liandra-giacomine (commented):
There's a failing test but otherwise, looks good

app/v2/controllers/V2MovementsController.scala
```diff
@@ -124,7 +129,22 @@ class V2MovementsControllerImpl @Inject() (
...
+    }
+
+  private def submitDepartureDeclarationLargeXML(): Action[Source[ByteString, _]] =
+    (authActionNewEnrolmentOnly andThen messageSizeAction()).stream {
+      implicit request =>
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+
+        (for {
+          upscan <- upscanService.upscanInitiate().asPresentation
+          _ = auditService.audit(AuditType.DeclarationData, request.body, MimeTypes.XML) // TODO - what data we need to send for auditing
```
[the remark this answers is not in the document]
prashanthmamidi: as discussed, commented out this code now

liandra-giacomine (approved):

PR #374, opened 2023-01-31
Ctcp 2003: Provide upload URL to trader initiating arrival 

liandra-giacomine:


app/v2/models/request/UpscanInitiate.scala
```diff
@@ -24,8 +24,5 @@ object UpscanInitiate {
 
 case class UpscanInitiate(
   callbackUrl: String,
-  successRedirect: Option[String],
```
prashanthmamidi, line 27: maybe we need these few of these optional fields later - if required we can later as well

test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -1545,82 +1595,229 @@ class V2MovementsControllerSpec
...
+      ) {
+        (upscanResponse, movementResponse, boxResponse) =>
+          beforeEach()
+
+          when(mockUpscanService.upscanInitiate(any[String].asInstanceOf[MovementId])(any(), any()))
+            .thenAnswer {
+              _ => EitherT.rightT(upscanResponse)
+            }
+
+          when(mockAuditService.audit(any(), any(), eqTo(MimeTypes.XML))(any(), any())).thenReturn(Future.successful(()))
```
prashanthmamidi, line 1705: we can remove this mock, as we are not sending any audit message

app/v2/controllers/V2MovementsController.scala
```diff
@@ -68,6 +67,7 @@ trait V2MovementsController {
   def getMovement(movementType: MovementType, movementId: MovementId): Action[AnyContent]
   def getMovements(movementType: MovementType, updatedSince: Option[OffsetDateTime], movementEORI: Option[EORINumber]): Action[AnyContent]
   def attachMessage(movementType: MovementType, movementId: MovementId): Action[Source[ByteString, _]]
+  def attachLargeMessage(movementId: MovementId, messageId: MessageId): Action[Source[ByteString, _]]
```
prashanthmamidi, line 70: tests are missing for attachLargeMessage

app/v2/models/request/UpscanInitiate.scala
liandra-giacomine: Yeah we might choose to use them later on, but since they are are always being set to None at the moment and are optional, I thought it was best to remove something we don't yet need

test/v2/controllers/V2MovementsControllerSpec.scala
liandra-giacomine: Good point!

app/v2/controllers/V2MovementsController.scala
liandra-giacomine: Will do

conf/application.conf
```diff
@@ -230,7 +230,13 @@ microservice {
...
       host = "localhost"
       port = 9570
-      maximumFileSize = 23068672  # 22 MB
+      maximumFileSize = 20000000  # 20 MB
+    }
+
+    common-transit-convention-traders {
+        protocol = "http"
+        host = "localhost"
+        port = 9487
```
drnaylor-oc, line 239: This is a bit recursive!

I wonder if there is a better way to get this -- I wonder what other services that need their own URL do. Might be worth a search.

liandra-giacomine: Agreed! I'll look into it

prashanthmamidi (approved):
LGTM

conf/application.conf
liandra-giacomine: As discussed, the only solution I found is less clean than this one :)

PR #375, opened 2023-02-02
[CTCP-2187] added audit for large message

prashanthmamidi:


app/v2/models/responses/LargeMessageAuditResponse.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import play.api.libs.json.Json
+import play.api.libs.json.OFormat
+import v2.models.MovementId
+import v2.models.MovementType
+
+case class LargeMessageAuditResponse(
+  movementId: MovementId,
+  movementType: MovementType,
+  xClientId: String,
```
liandra-giacomine, line 27: You could keep xClientId as Option[String] rather than passing an empty string if it's not present

drnaylor-oc: indeed, in fact I would much rather that.

derekspalmer (approved):

PR #380, opened 2023-02-10
CTCP-2402: Added status to message response from transit-movements

liandra-giacomine:


app/v2/controllers/V2MovementsController.scala
```diff
@@ -330,10 +330,8 @@ class V2MovementsControllerImpl @Inject() (
     }
   }
 
-  def attachLargeMessage(movementId: MovementId): Action[Source[ByteString, _]] = authActionNewEnrolmentOnly.async(streamFromMemory) {
+  def attachLargeMessage(movementId: MovementId): Action[JsValue] = authActionNewEnrolmentOnly.async(parse.json) {
     implicit request =>
```
derekspalmer, line 334: Is this actually attaching anything?

derekspalmer (approved):

app/v2/controllers/V2MovementsController.scala
liandra-giacomine: Not yet, we will be doing that in the upcoming sprints

PR #383, opened 2023-02-14
[CTCP-2403] added support to return messageId for large messages

prashanthmamidi:


app/v2/controllers/V2MovementsController.scala
```diff
@@ -198,10 +198,16 @@ class V2MovementsControllerImpl @Inject() (
 
         (for {
           movementResponse  <- movementsService.createMovement(request.eoriNumber, movementType, None).asPresentation
-          upscanResponse    <- upscanService.upscanInitiate(movementResponse.movementId).asPresentation
+          upscanResponse    <- upscanService.upscanInitiate(movementResponse.movementId, movementResponse.messageId).asPresentation
           boxResponseOption <- mapToBoxResponse(pushNotificationsService.associate(movementResponse.movementId, movementType, request.headers))
           auditResponse = Json.toJson(
-            LargeMessageAuditResponse(movementResponse.movementId, movementType, request.headers.get(XClientIdHeader), upscanResponse)
+            LargeMessageAuditResponse(
```
liandra-giacomine, line 204: LargeMessageAuditResponse - this isn't a response though, it's a request? Should it not be LargeMessageAuditRequest?

app/v2/services/MovementsService.scala
```diff
@@ -88,12 +88,7 @@ class MovementsServiceImpl @Inject() (persistenceConnector: PersistenceConnector
       persistenceConnector
         .postMovement(eori, movementType, source)
         .map {
-          movementResponse =>
-            if (movementResponse.messageId.isEmpty && source.isDefined) { // For Small message, messageId shouldn't be empty
-              Left(PersistenceError.MessageIdError())
```
liandra-giacomine, line 93: PersistenceError.MessageIdError can be removed

app/v2/controllers/V2MovementsController.scala
```diff
@@ -198,10 +198,16 @@ class V2MovementsControllerImpl @Inject() (
 
         (for {
           movementResponse  <- movementsService.createMovement(request.eoriNumber, movementType, None).asPresentation
-          upscanResponse    <- upscanService.upscanInitiate(movementResponse.movementId).asPresentation
+          upscanResponse    <- upscanService.upscanInitiate(movementResponse.movementId, movementResponse.messageId).asPresentation
           boxResponseOption <- mapToBoxResponse(pushNotificationsService.associate(movementResponse.movementId, movementType, request.headers))
```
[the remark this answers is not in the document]
prashanthmamidi: we need upscanResponse to send this back to the trader and to the audit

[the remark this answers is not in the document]
liandra-giacomine: yes, I'm talking about the PPNS call so:

`boxResponseOption =`

conf/app.routes
```diff
@@ -23,4 +23,4 @@ GET        /movements/departures/:departureId/messages/:messageId    routing.Dep
 
 GET        /movements/push-pull-notifications/box                   controllers.PushPullNotificationController.getBoxInfo()
 
-POST       /movements/:movementId/messages                          v2.controllers.V2MovementsController.attachLargeMessage(movementId: v2.models.MovementId)
+POST       /movements/:movementId/messages/:messageId                         v2.controllers.V2MovementsController.attachLargeMessage(movementId: v2.models.MovementId, messageId: v2.models.MessageId)
```
sonisn1988, line 26: @drnaylor-oc  / @liandra-giacomine  I think here we need to define two type of call back ? one for arrival and another for departure like /movements/arrivals/:movementId/messages/:messageId . As later when we will verify messageType on backend we would not be able to differentiate between arrival/departure messageType with same callback url (i.e. /movements/:movementId/messages/:messageId). so there are possibilities trader can pass departure messageType or arrival endpoint or visa versa ....

drnaylor-oc: Actually, that's a good question. We could either update the URL to include that, or we could query the database instead.

The complication here is that we might need more information -- for example, are we expecting a IE015 because one hasn't been successfully sent yet? Maybe it should be anything but an IE015. How do we do this?

Something we need to think about -- but maybe not for this PR because in true agile style, we want to get the minimum viable solution, and then adapt that as we build on it.

Worth a spike or at least a discussion.

app/v2/controllers/V2MovementsController.scala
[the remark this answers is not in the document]
prashanthmamidi: ok, I think it's same we have for small message calls as well...we have handled as below, where we return None instead o failing for any error `  private def mapToBoxResponse(
    boxResponseEither: EitherT[Future, PushNotificationError, BoxResponse]
  ): EitherT[Future, PresentationError, Option[BoxResponse]] =
    EitherT[Future, PresentationError, Option[BoxResponse]] {
      boxResponseEither.fold(
        _ => Right(None),
        r => Right(Some(r))
      )
    }`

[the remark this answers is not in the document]
liandra-giacomine: Oh right yeah! Sorry! :)

test/v2/services/MovementsServiceSpec.scala
```diff
@@ -122,9 +113,9 @@ class MovementsServiceSpec
     "on a successful submission, should return a Right" in {
       implicit val hc: HeaderCarrier = HeaderCarrier()
       when(mockConnector.postMovement(EORINumber(any[String]), any(), any())(eqTo(hc), any[ExecutionContext]))
-        .thenReturn(Future.successful(MovementResponse(MovementId("ABC"), None)))
+        .thenReturn(Future.successful(MovementResponse(MovementId("ABC"), MessageId("123"))))
```
sonisn1988, line 116: movementID and messageId should be 16 digit for version 2 api. Good if we can test with right data otherwise please ignore

app/v2/models/responses/hateoas/HateoasMovementMessageIdsResponse.scala
```diff
@@ -32,20 +33,31 @@ object HateoasMovementMessageIdsResponse extends HateoasResponse {
...
-            s"${movementType.movementType}Id" -> movementId.value,
-            "received"                        -> message.received,
-            "type"                            -> message.messageType,
-            "status"                          -> message.status
+          val jsMessageLinksObject = Json.obj(
+            "self"                    -> Json.obj("href" -> getMessageUri(movementId, message.id, movementType)),
+            movementType.movementType -> Json.obj("href" -> getMovementUri(movementId, movementType))
           )
-      )
+          if (message.status == MessageStatus.Pending) {
```
sonisn1988, line 42: I am not in the favour of hiding something that is available in database. Its look conflict with me. As if someone is checking with database they will find messageType but when they look in response they would not find it for large messageType. So as I commented on transit-movement backend api we should do this as part of callback. Its my suggestion. But if you guys think that is fine then no problem.

drnaylor-oc: The trader is _never_ going to be able to check the database. Remember the role of this service in this scenario is to create the view model, which does not, and frankly, should not, be driven by the database model.

I've not checked the code itself so I don't have an opinion of whether this is the "right" way to hide it, but it's perfectly fine for it to be done here.

liandra-giacomine: If the type is pending, we won't know the message type until we get the message - the status will change from Pending and therefore we will return the  message type

app/v2/models/responses/hateoas/HateoasMovementMessageIdsResponse.scala
```diff
@@ -32,20 +33,31 @@ object HateoasMovementMessageIdsResponse extends HateoasResponse {
...
+              s"${movementType.movementType}Id" -> movementId.value,
+              "received"                        -> message.received,
+              "status"                          -> message.status
+            )
+          } else {
+            Json.obj(
+              "_links"                          -> jsMessageLinksObject,
+              "id"                              -> message.id.value,
+              s"${movementType.movementType}Id" -> movementId.value,
+              "received"                        -> message.received,
```
liandra-giacomine, line 55: Message type is optional right? We could prevent the duplicated code by doing something similar to what we do for body in the Hateos response for the movement response

```
  message.messageType .map(
         type => Json.obj("type" -> type)
       )
       .getOrElse(Json.obj())
```

prashanthmamidi: nope, Message type is not optional

test/v2/services/MovementsServiceSpec.scala
prashanthmamidi: it's a test data, it should be fine

app/v2/models/responses/hateoas/HateoasMovementMessageIdsResponse.scala
drnaylor-oc: hmmm, that's gonna have to change in the future for reasons given above -- but for now it can probably be left as is because we've not hit the point where it _has_ to be optional and I don't want scope creep to accidentally occur.

test/v2/services/MovementsServiceSpec.scala
drnaylor-oc: I would rather test data reflect real world cases as much as possible when it affects the test -- but I'm as guilty as others for doing this similar thing.

Would rather it was at least a 16 character hex, and maybe it should use the generators, but considering what this test is, I'd not get hung up on that.

prashanthmamidi: anyways, updated to use 16 digit to reflect real world cases

sonisn1988 (approved):

liandra-giacomine (approved):

PR #384, opened 2023-02-17
added support for version2 accept headers

akila-sankar:


app/routing/ArrivalsRouter.scala
```diff
@@ -50,17 +50,15 @@ class ArrivalsRouter @Inject() (
     with StreamingParsers
     with VersionedRouting {
 
-  val VERSION_2_ACCEPT_HEADER_PATTERN = """^application\/vnd\.hmrc\.2\.0\+.+$""".r
-
   def createArrivalNotification(): Action[Source[ByteString, _]] =
     route {
-      case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON) => v2Arrivals.createMovement(MovementType.Arrival)
-      case _                                                         => v1Arrivals.createArrivalNotification()
+      case Some(VersionedRouting.VERSION_2_ACCEPT_HEADER_PATTERN()) => v2Arrivals.createMovement(MovementType.Arrival)
```
liandra-giacomine, line 55: For all endpoints besides the Get single message endpoint, we return the JSON so the only accept header we should allow is this one VERSION_2_ACCEPT_HEADER_VALUE_JSON. 

For the Get single message endpoint, the user can specify whether they want their response in json with the message in XML, or just in JSON where we convert their XML message to JSON.

This isn't the case for all other endpoints since all the data returned is solely json. This change here is correct, since we want to route to version 2 if they have used the version 2 pattern for the accept header, but we need more refinement in the action to determine whether the endpoint accepts JSON or also JSON wrapped XML

app/v2/controllers/V2MovementsController.scala
```diff
@@ -122,7 +123,7 @@ class V2MovementsControllerImpl @Inject() (
     }
 
   private def submitDepartureDeclarationXML(): Action[Source[ByteString, _]] =
-    (authActionNewEnrolmentOnly andThen messageSizeAction()).stream {
+    (authActionNewEnrolmentOnly andThen acceptHeaderActionProvider(isOnlyJson) andThen messageSizeAction()).stream {
```
liandra-giacomine, line 126: instead of setting a val to isOnlyJson and passing it or !isOnlyJson, can we just pass the boolean through instead please?

you can specify the parameter in the call to make what you're passing more intuitive:

`acceptHeaderActionProvider(isOnlyJson = true)`

akila-sankar: Updated with the change mentioned above

app/v2/controllers/actions/AcceptHeaderAction.scala
```diff
@@ -55,7 +56,13 @@ class AcceptHeaderActionImpl[R[_] <: Request[_]] @Inject() (implicit val executi
     }
 
   private def checkAcceptHeader[A](acceptHeaderValue: String, request: R[A]): Either[Result, R[A]] =
-    if (acceptedHeaders.contains(acceptHeaderValue)) Right(request)
-    else Left(NotAcceptable(Json.toJson(PresentationError.notAcceptableError(s"Accept header $acceptHeaderValue is not supported!"))))
+    if (
+      (!isOnlyJson && acceptedHeaders
+        .contains(acceptHeaderValue)) || (isOnlyJson && VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON.contains(acceptHeaderValue))
+    ) {
```
drnaylor-oc, line 62: This feels a bit messy to me -- once an if statement goes beyond two conditions it's worth starting to think about how this might be done better.

The goal of the `isOnlyJson` boolean is to limit the accepted headers to the Json header. So, instead of making this change here and doing this conditional every time, I suggest reverting this method back to what it was before, and changing the `acceptedHeaders` lazy val, like so:

```scala
private lazy val acceptedHeaders = 
    if (isOnlyJson) Seq(VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON) 
    else Seq(
        VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON,
        VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON_XML,
        VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON_XML_HYPHEN
    )
```

That way we initialise the `acceptedHeaders` once, store the list, and the `checkAcceptHeader` function can continue to check this seq.

app/v2/controllers/actions/AcceptHeaderAction.scala
```diff
@@ -31,7 +31,8 @@ import scala.concurrent.Future
 
 trait AcceptHeaderAction[R[_] <: Request[_]] extends ActionRefiner[R, R]
 
-class AcceptHeaderActionImpl[R[_] <: Request[_]] @Inject() (implicit val executionContext: ExecutionContext) extends AcceptHeaderAction[R] {
+class AcceptHeaderActionImpl[R[_] <: Request[_]] @Inject() (isOnlyJson: Boolean)(implicit val executionContext: ExecutionContext)
```
drnaylor-oc, line 34: It might be better to rename the parameter `acceptOnlyJson`, because that's more indicative of what we're doing.

app/v2/controllers/actions/providers/AcceptHeaderActionProvider.scala
```diff
@@ -27,12 +27,12 @@ import scala.concurrent.ExecutionContext
...
 
-  def apply[R[_] <: Request[_]](): AcceptHeaderAction[R]
+  def apply[R[_] <: Request[_]](isOnlyJson: Boolean): AcceptHeaderAction[R]
 
 }
 
 class AcceptHeaderActionProviderImpl @Inject() ()(implicit ec: ExecutionContext) extends AcceptHeaderActionProvider {
 
-  def apply[R[_] <: Request[_]](): AcceptHeaderAction[R] = new AcceptHeaderActionImpl[R]
+  def apply[R[_] <: Request[_]](isOnlyJson: Boolean): AcceptHeaderAction[R] = new AcceptHeaderActionImpl[R](isOnlyJson)
```
drnaylor-oc, line 36: Same here -- this should also be `acceptOnlyJson`.

app/v2/controllers/actions/AcceptHeaderAction.scala
```diff
@@ -55,7 +56,13 @@ class AcceptHeaderActionImpl[R[_] <: Request[_]] @Inject() (implicit val executi
...
   private def checkAcceptHeader[A](acceptHeaderValue: String, request: R[A]): Either[Result, R[A]] =
-    if (acceptedHeaders.contains(acceptHeaderValue)) Right(request)
-    else Left(NotAcceptable(Json.toJson(PresentationError.notAcceptableError(s"Accept header $acceptHeaderValue is not supported!"))))
+    if (
+      (!isOnlyJson && acceptedHeaders
+        .contains(acceptHeaderValue)) || (isOnlyJson && VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON.contains(acceptHeaderValue))
+    ) {
+      Right(request)
+    } else {
+      Left(NotAcceptable(Json.toJson(PresentationError.notAcceptableError(s"Accept header $acceptHeaderValue is not supported!"))))
```
drnaylor-oc, line 65: Can you change this message to `The Accept header is missing or invalid.` please, to bring it in line with comments on https://github.com/hmrc/common-transit-convention-traders/pull/386#pullrequestreview-1307583448?

app/routing/VersionedRouting.scala
```diff
@@ -68,9 +69,9 @@ trait VersionedRouting {
             // To avoid a memory leak, we need to ensure we run the request stream and ignore it.
             request.body.to(Sink.ignore).run()
             Future.successful(
-              UnsupportedMediaType(
+              NotAcceptable(
                 Json.toJson(
-                  PresentationError.unsupportedMediaTypeError(
+                  PresentationError.notAcceptableError(
```
drnaylor-oc, line 74: Similarly, I realise this is out of your changeset, but the lines below (75 to 80), can you delete them, and just return "The Accept header is missing or invalid."?

app/v2/controllers/actions/AcceptHeaderAction.scala
akila-sankar: Yes updated

app/v2/controllers/actions/AcceptHeaderAction.scala
akila-sankar: Sure updated

app/routing/VersionedRouting.scala
akila-sankar: Ideally the scenario is for 406 instead of 413, so corrected it & removed lines 75 to 80 and updated the error desc to above text

drnaylor-oc (approved):
Looks okay to me, but please don't merge until https://github.com/hmrc/ctc-traders-api-tests/pull/86 is approved and merged -- the accept header naming broke the regression test pack slightly.

akila-sankar:
Apologies, had an issue with GPG key and created another PR (https://github.com/hmrc/common-transit-convention-traders/pull/388) with same changes and closing this

PR #385, opened 2023-02-20
Added status to single message and message ids endpoints

liandra-giacomine:


fmcgrath (commented):
I wonder if the API definition needs to explain the purpose/context of the `status` parameter. It might not be sufficient to explain it in the API changelog.

resources/public/api/conf/2.0/application.yaml
```diff
@@ -275,7 +275,15 @@ paths:
...
-
+        
+        Below is the set of all possible values returned in the message status field and their meaning:
+        <ul>
+          <li>Pending: the file has not yet been uploaded or we’re waiting for checks to complete</li>
+          <li>Processing: the file has been received by us and we’re processing the message</li>
+          <li>Success: the file has been processed successfully</li>
+          <li>Failed: there was an issue with the file and the message was not processed</li>
+        </ul>
+
```
fmcgrath, line 286: This is useful but a few minor problems:

- The intro is longer than it needs to be.
- HTML tags are not needed - Markdown can suffice.
- GDS style guide advises against using contractions, such as "we're".
- status should have back ticks to highlight the fact that it is a parameter name.

Suggestion:
The `status` parameter can have the following values:

- Pending: the file has not yet been uploaded or checks are not yet completed
- Processing: the file has been received by us and the message is being processed
- Success: the file has been processed successfully
- Failed: there was a problem with the file and the message was not processed

resources/public/api/conf/2.0/application.yaml
```diff
@@ -473,6 +481,14 @@ paths:
         
         * `application/vnd.hmrc.2.0+json`: JSON, where the message is provided as a sub-document sent as part of a JSON payload containing other metadata.
         
+        Below is the set of all possible values returned in the message status field and their meaning:
+          <ul>
+            <li>Pending: the file has not yet been uploaded or we’re waiting for checks to complete</li>
+            <li>Processing: the file has been received by us and we’re processing the message</li>
+            <li>Success: the file has been processed successfully</li>
+            <li>Failed: there was an issue with the file and the message was not processed</li>
+          </ul>
```
fmcgrath, line 490: See comment above.

fmcgrath (changes requested):
Text is good but needs some minor tweaks.

fmcgrath (approved):
Looks good.

PR #386, opened 2023-02-21
CTCP-2355 version2 accept headers 406 api definition update

akila-sankar:


derekspalmer (dismissed):

resources/public/api/conf/2.0/application.yaml
```diff
@@ -70,6 +70,20 @@ paths:
...
+              schema:
+                allOf:
+                  - $ref: '#/components/schemas/AcceptHeaderInvalid'
+                  - description: Accept header application/vnd.hmrc.2.0+json123 is not supported!.
+                    example:
+                      code: NOT_ACCEPTABLE
+                      message: Accept header application/vnd.hmrc.2.0+json123 is not supported!
+              example:
+                code: NOT_ACCEPTABLE
+                message: Accept header application/vnd.hmrc.2.0+json123 is not supported!
```
fmcgrath, line 86: A few issues:

- Up to now, the preferred `description` for a `406` was 'The Accept header is missing or invalid.'
- Why write` json123`?
- Why use an exclamation mark (`!`)?

resources/public/api/conf/2.0/application.yaml
```diff
@@ -166,13 +180,13 @@ paths:
...
                     example:
-                      statusCode: 406
-                      message: The accept header is missing or invalid
+                      code: NOT_ACCEPTABLE
+                      message: Accept header application/vnd.hmrc.2.0+json123 is not supported!
               example:
-                statusCode: 406
-                message: The accept header is missing or invalid
+                code: NOT_ACCEPTABLE
+                message: Accept header application/vnd.hmrc.2.0+json123 is not supported!
```
fmcgrath, line 189: See comment above.

resources/public/api/conf/2.0/application.yaml
```diff
@@ -1181,14 +1237,16 @@ components:
           description: The description of the error.
     AcceptHeaderInvalid:
       title: AcceptHeaderInvalid
-      description: The `Accept` header was not set to `application/vnd.hmrc.2.0+json`.
+      description: The `Accept` header was not set to `application/vnd.hmrc.2.0+json`, `application/vnd.hmrc.2.0+json+xml` or `application/vnd.hmrc.2.0+json-xml`.
```
fmcgrath, line 1240: Was there not some Platform problem with writing `json+xml` in API source code and API definitions? That's why we went with `json-xml`.

liandra-giacomine, line 1240: The platform doesn't allow application/vnd.hmrc.2.0+json-xml, please remove

resources/public/api/conf/2.0/application.yaml
```diff
@@ -1181,14 +1237,16 @@ components:
           description: The description of the error.
     AcceptHeaderInvalid:
       title: AcceptHeaderInvalid
-      description: The `Accept` header was not set to `application/vnd.hmrc.2.0+json`.
+      description: The `Accept` header was not set to `application/vnd.hmrc.2.0+json`, `application/vnd.hmrc.2.0+json+xml` or `application/vnd.hmrc.2.0+json-xml`.
       required:
-        - statusCode
+        - code
```
liandra-giacomine, line 1242: Why was this change made?

resources/public/api/conf/2.0/application.yaml
drnaylor-oc: Yeah. We left it in the code in case platform allowed it at some point -- so it's not entirely obvious that's the case just by looking at it.

@akila-sankar Can you remove `application/vnd.hmrc.2.0+json+xml` from the list please?

resources/public/api/conf/2.0/application.yaml
drnaylor-oc: To be fair, that was already in the code like that from before. I imagine this was masked because API Platform overrides the messages for many of the codes.

I don't disagree with what you said though, we should fix the messages in our code to fit the style of all other messages, platform provided or otherwise.

resources/public/api/conf/2.0/application.yaml
akila-sankar: Yes agree @drnaylor-oc will remove from the list

resources/public/api/conf/2.0/application.yaml
drnaylor-oc: "The Accept header is missing or invalid."

fmcgrath (dismissed):
Looks good.

drnaylor-oc (approved):
Same as https://github.com/hmrc/common-transit-convention-traders/pull/384#pullrequestreview-1308759730 -- merge at the same time as that.

akila-sankar:
Apologies had an issue with GPG key & created another PR(https://github.com/hmrc/common-transit-convention-traders/pull/387) with same changes and closing this.

PR #390, opened 2023-02-23
Ctcp 2016

liandra-giacomine:


app/v2/services/ObjectStoreService.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  private val dateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMdd-HHmmss").withZone(ZoneOffset.UTC)
+
+  override def addMessage(downloadUrl: DownloadUrl, movementId: MovementId, messageId: MessageId)(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): EitherT[Future, ObjectStoreError, ObjectSummaryWithMd5] =
+    EitherT {
+      val formattedDateTime = dateTimeFormatter.format(OffsetDateTime.ofInstant(clock.instant, ZoneOffset.UTC))
+
```
liandra-giacomine, line 65: @drnaylor-oc This feels really imperative, is there a more functional way to catch the exception that gets thrown when an invalid URL is passed here `new URL(downloadUrl.value)`?

liandra-giacomine: Thanks!

app/v2/services/ObjectStoreService.scala
```diff
@@ -0,0 +1,81 @@
...
+      (for {
+        url <- Future.fromTry(Try(new URL(downloadUrl.value)))
+        response <- client
+          .uploadFromUrl(
+            from = url,
+            to = Path.Directory("common-transit-convention-traders").file(s"${movementId.value}-${messageId.value}-$formattedDateTime.xml")
```
sonisn1988, line 70: Please "common-transit-convention-traders" from here as that will be added automatically while generating object store uri.
Upload object from url
osClient.uploadFromUrl(
  from            = new URL("https://fus-outbound.s3.eu-west-2.amazonaws.com/81fb03f5-195d-422a-91ab-460939045846"),
  to              = Path.File(Path.Directory("my-folder"), "sample.pdf"),
  retentionPeriod = RetentionPeriod.OneWeek, // defaults to 'object-store.default-retention-period' configuration
  contentType     = Some("text/csv"), // defaults to None which results in the contentType 'application/octet-stream'
  contentMd5      = Some(Md5Hash("a3c2f1e38701bd2c7b54ebd7b1cd0dbc")), // defaults to None
  owner           = "my-service" // defaults to 'appName' configuration
) // returns Future[ObjectSummaryWithMd5]
The above code will download a file from https://fus-outbound.s3.eu-west-2.amazonaws.com/81fb03f5-195d-422a-91ab-460939045846 to **/my-service/**my-folder/sample.pdf **my-service** is going to be **common-transit-convention-traders** for our case.

drnaylor-oc: To make it clearer:

`to` should be to `movements/<movementid>/<filename>`

`owner` should be set to `common-transit-convention-traders` (that would be the default in this case, but it will  make it clear that's the owning service).

drnaylor-oc: Also, the retention period should be 7 years.

liandra-giacomine: @drnaylor-oc @sonisn1988  The `to`  parameter takes in the following case class. So would the directory be `movements/<movementid>`? 

 case class File(directory: Directory, fileName: String)

drnaylor-oc: Yes

liandra-giacomine: Thanks guys

VijayKumarTech:
Tested this against local - able to retrieve file from Object Store url.
Happy to approve.

VijayKumarTech (dismissed):
Tested this against local - able to retrieve file from Object Store url.
Happy to Merge upon completion of peer review.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -338,10 +339,21 @@ class V2MovementsControllerImpl @Inject() (
...
+          .map {
+            upscanResponse =>
+              (for {
+                objectStoreResponse <- objectStoreService.addMessage(upscanResponse.downloadUrl.get, movementId, messageId).asPresentation
+              } yield objectStoreResponse).fold[Result](
+                _ => Ok, //TODO: Send notification to PPNS with details of the error
+                _ => Ok  //TODO: Send notification to PPNS with details of the success
+              )
+          }
+          .fold[Result](
```
drnaylor-oc, line 353: A bit nitpicky, but it took me a moment to understand why you're `map`ping here rather than parsing the upscan response in the for comprehension -- I get it now in that if Upscan sends rubbish, we need to tell them as such.

However, if we have a failed `UpscanResponse`, this code is going to fall over because of `upscanResponse.downloadUrl.get`. That needs to be handled.

liandra-giacomine: In parseAndLogUpscanResponse we ensure that we have the download URL for a successful callback, otherwise we send them 400 - that's something I spotted in Prasanth's PR and it needs to also be updated in the router.

liandra-giacomine: My bad, you're right!

test/v2/controllers/ErrorTranslatorSpec.scala
```diff
@@ -26,7 +26,7 @@ import org.scalatest.freespec.AnyFreeSpec
 import org.scalatest.matchers.must.Matchers
```
sonisn1988, line 26: Nitpick -- as you have changes the class name to ``` ConvertError Spec ```  test class should be ``` ConvertErrorSpec ``` .

test/v2/fakes/objectstore/ObjectStoreStub.scala
```diff
@@ -0,0 +1,52 @@
...
+  override def uploadFromUrl(
+    from: java.net.URL,
+    to: Path.File,
+    retentionPeriod: RetentionPeriod = config.defaultRetentionPeriod,
+    contentType: Option[String] = None,
+    contentMd5: Option[Md5Hash] = None,
+    owner: String = config.owner
+  )(implicit hc: HeaderCarrier): FutureEither[ObjectSummaryWithMd5] = {
+    val objectSummaryWithMd5 = arbitraryObjectSummaryWithMd5.arbitrary.sample.get
+    Future.successful(Right(objectSummaryWithMd5))
```
sonisn1988, line 49: Need to define stub for failed future and write test case in ObjectStoreServiceSpec for failed future for unexpected error at Object store while copying file.

liandra-giacomine: Not sure how that can be done with the stub

app/v2/controllers/UpscanResponseParser.scala
```diff
@@ -16,43 +16,38 @@
...
       case UpscanResponse(reference, _, Some(_), Some(_), None) =>
         logger.info(s"Received a successful response from Upscan callback for the following reference: $reference")
-        Future.successful(Right(upscanResponse))
+        Right(upscanResponse)
       case UpscanResponse(reference, _, None, None, Some(failureDetails)) =>
         logger.warn(
           s"Received a failure response from Upscan callback for the following reference: $reference. Failure reason: ${failureDetails.failureReason}. Failure message: ${failureDetails.message}"
         )
-        Future.successful(Right(upscanResponse))
+        Right(upscanResponse)
```
sonisn1988, line 48: I am little bit confused with this case condition. so In case of Upscan response with failureDetails . should we not return 500 or Bad request or some error and should not proceed further. As looks like in case of Upscan response with failureDetails it is going to proceed further and then making call to Object store to copy the file from upscan with download url but we would not be getting download url in case of Upscan response with failureDetails? Please correct me if I misunderstood.

liandra-giacomine: If we can't parse the request from upscan, we return a bad request. 
If we can parse the request from upscan, it could be a failure or a success response. We handle that in handleUpscanSuccessResponse, where if the download URL is not present, we will be updating the trader about it when we have implemented PPNS, and returning 200. 

```
private def handleUpscanSuccessResponse(upscanResponse: UpscanResponse): EitherT[Future, PresentationError, DownloadUrl] =
     EitherT {
       Future.successful(upscanResponse.downloadUrl.toRight {
         PresentationError.badRequestError("Upscan failed to process file")
       })
     }
```

This ensures we have a downloadUrl and if we don't we return a presentation error so it exists out of the for comprehension with that. Hope that makes sense

sonisn1988: ok my bad. thanks.

sonisn1988 (approved):
LGTM

PR #391, opened 2023-02-24
[CTCP-2580] add endpoint to support callback from upscan

prashanthmamidi:


app/v2/controllers/UpscanResponseParser.scala
```diff
@@ -0,0 +1,59 @@
...
+            Future.successful(Right(upscanResponse))
+        }
+        .getOrElse {
+          logResponse(None)
+          Future.successful(Left(PresentationError.badRequestError("Unexpected Upscan callback response")))
+        }
+    }
+  }
+
+  private def logResponse(upscanResponse: Option[UpscanResponse]) =
```
drnaylor-oc, line 48: I suppose this works, but it's a bit weird that you're handling code from two different branches, that'll do two different things, in this function -- you'll only get `None` from the `getOrElse` part of the function above, you'll only get `Some` from the `map` part. 

I think this should just accept a `UpscanResponse` and deal with success cases. The failure case (`None`) is a simple one-liner that should go in the appropriate place in the function above -- it's a little less complicated that way.

It also means you don't have to re-wrap the argument in `map` with a `Some`.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -337,9 +338,12 @@ class V2MovementsControllerImpl @Inject() (
   }
 
   def attachLargeMessage(movementId: MovementId, messageId: MessageId): Action[JsValue] =
-    (authActionNewEnrolmentOnly andThen acceptHeaderActionProvider(acceptOnlyJson = true)).async(parse.json) {
+    Action.async(parse.json) {
```
liandra-giacomine, line 341: @drnaylor-oc We should still authenticate these requests right and only accept it if coming from upscan (what if a trader manages to guess the endpoint for example) - not sure what the best way to go about it is, perhaps a spike to investigate?

drnaylor-oc: If we can authenticate it, I'd want it, very much so -- if only to ensure we know we're talking to Upscan.

However, the trader won't be able to get here because it's not in the API definition. That definition controls the endpoints that the API platform proxy lets through. So, as long as we don't put it in the API definition, that's not a worry.

liandra-giacomine: Oh, that's good to know! Thanks :)

app/v2/controllers/UpscanResponseParser.scala
prashanthmamidi: yes, it's looks much simpler now.

liandra-giacomine (approved):

PR #394, opened 2023-03-07
CTCP-2517: Show how we can extract the file size

drnaylor-oc:
**Not for merging**. Just here to show how we can do this.


drnaylor-oc:
This has served its purpose.

PR #396, opened 2023-03-13
CTCP-2435

sonisn1988:
Call router from API layer (common-transit-convention-traders)

### DIFF
--- a/app/v2/controllers/V2MovementsController.scala
+++ b/app/v2/controllers/V2MovementsController.scala
@@ -352,8 +352,19 @@ class V2MovementsControllerImpl @Inject() (
               downloadUrl   <- handleUpscanSuccessResponse(upscanResponse)
               objectSummary <- objectStoreService.addMessage(downloadUrl, movementId, messageId).asPresentation
               messageUpdate = MessageUpdate(MessageStatus.Processing, Some(ObjectStoreURI(objectSummary.location.asUri)))
-              update <- persistenceService.updateMessage(movementId, messageId, messageUpdate).asPresentation
-            } yield update).fold[Result](
+
+              _ <- persistenceService.updateMessage(movementId, messageId, messageUpdate).asPresentation
+
+              sendMessage <- routerService
+                .sendLargeMessage(
+                  MessageType.DeclarationData, // Only IE015 is supported as of now for large message
+                  EORINumber("123456789"),     // Eori number ??
+                  movementId,
+                  messageId,
+                  ObjectStoreURI(objectSummary.location.asUri)
+                )
+                .asPresentation
+            } yield sendMessage).fold[Result](
               _ => Ok, //TODO: Send notification to PPNS with details of the error
               _ => Ok  //TODO: Send notification to PPNS with details of the success
             )
--- a/it/v2/connectors/RouterConnectorSpec.scala
+++ b/it/v2/connectors/RouterConnectorSpec.scala
@@ -19,10 +19,7 @@ package v2.connectors
 import akka.stream.Materializer
 import akka.stream.scaladsl.Source
 import akka.util.ByteString
-import com.github.tomakehurst.wiremock.client.WireMock.aResponse
-import com.github.tomakehurst.wiremock.client.WireMock.equalTo
-import com.github.tomakehurst.wiremock.client.WireMock.post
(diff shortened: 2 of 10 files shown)


app/v2/controllers/V2MovementsController.scala
```diff
@@ -352,8 +352,19 @@ class V2MovementsControllerImpl @Inject() (
...
               objectSummary <- objectStoreService.addMessage(downloadUrl, movementId, messageId).asPresentation
               messageUpdate = MessageUpdate(MessageStatus.Processing, Some(ObjectStoreURI(objectSummary.location.asUri)))
-              update <- persistenceService.updateMessage(movementId, messageId, messageUpdate).asPresentation
-            } yield update).fold[Result](
+
+              _ <- persistenceService.updateMessage(movementId, messageId, messageUpdate).asPresentation
+
+              sendMessage <- routerService
+                .sendLargeMessage(
+                  MessageType.DeclarationData, // Only IE015 is supported as of now for large message
```
liandra-giacomine, line 360: we should be extracting the message type here, if it's not a MessageType.DeclarationData then we can fail the request - I think it would be wrong to hard-code it here since it is technically possible to not send that specific type

it/v2/connectors/RouterConnectorSpec.scala
```diff
@@ -75,31 +72,29 @@ class RouterConnectorSpec
   implicit lazy val ec: ExecutionContext       = app.materializer.executionContext
   lazy val routerConnector: RouterConnector    = new RouterConnectorImpl(new TestMetrics(), appConfig, httpClientV2)
 
-  "POST /traders/:eori/message/:movementType/:messageId/movements/:movementId" - {
+  def targetUrl(eoriNumber: EORINumber, messageType: MessageType, movementId: MovementId, messageId: MessageId) =
+    s"/transit-movements-router/traders/${eoriNumber.value}/movements/${messageType.movementType.urlFragment}/${movementId.value}/messages/${messageId.value}"
 
-    lazy val messageTypeGen = Gen.oneOf(Seq(MessageType.DeclarationData))
+  implicit val hc = HeaderCarrier()
```
liandra-giacomine, line 78: implicit header carrier?!

liandra-giacomine (commented):
looking good overall, just a comment regarding extracting the message type

app/v2/controllers/V2MovementsController.scala
sonisn1988: @drnaylor-oc / @liandra-giacomine  Not sure if I misunderstood, but in morning call Dan said that we can hardcode ``` MessageType.DeclarationData ``` as if you can see for createDepartureDeclaration we are not validating MessageType. just simple passing that  ``` MessageType.DeclarationData  ```

if we want to validate MessageType here then I think we need to make get call to object store get the source and then extract the messageType using existing logic  ``` xmlParsingService.extractMessageType(source, messageTypeList).asPresentation ```  -- messageTypeList = Seq(MessageType.DeclarationData)

is that what we expecting here?

liandra-giacomine: Oh fair enough, we can do that in a later ticket then

app/v2/controllers/V2MovementsController.scala
```diff
@@ -351,15 +346,32 @@ class V2MovementsControllerImpl @Inject() (
             (for {
               downloadUrl   <- handleUpscanSuccessResponse(upscanResponse)
               objectSummary <- objectStoreService.addMessage(downloadUrl, movementId, messageId).asPresentation
+
               messageUpdate = MessageUpdate(MessageStatus.Processing, Some(ObjectStoreURI(objectSummary.location.asUri)))
-              update <- persistenceService.updateMessage(movementId, messageId, messageUpdate).asPresentation
-            } yield update).fold[Result](
+              _ <- persistenceService.updateMessage(eori, movementType, movementId, messageId, messageUpdate).asPresentation
+              messageType = extractMessageType(movementType)
```
drnaylor-oc, line 352: Can we add a TODO here to properly extract the message type later? I get why we've not done it now, but it can't be forgotten.

olakunlekuye2 (approved):

drnaylor-oc (approved):
Alright, let's get this show on the road.

PR #397, opened 2023-03-13
CTCP-2376 & CTCP-2375

liandra-giacomine:


conf/app.routes
```diff
@@ -1,26 +1,26 @@
...
+POST       /movements/departures/:departureId/messages                      routing.DeparturesRouter.attachMessage(departureId: String)
+GET        /movements/departures/:departureId/messages                      routing.DeparturesRouter.getMessageIds(departureId: String, receivedSince: Option[OffsetDateTime] ?= None)
 
-GET        /movements/departures/:departureId                       routing.DeparturesRouter.getDeparture(departureId: String)
+GET        /movements/departures/:departureId/messages/:messageId           routing.DeparturesRouter.getMessage(departureId: String, messageId: String)
 
-POST       /movements/departures/:departureId/messages              routing.DeparturesRouter.attachMessage(departureId: String)
-GET        /movements/departures/:departureId/messages              routing.DeparturesRouter.getMessageIds(departureId: String, receivedSince: Option[OffsetDateTime] ?= None)
+GET        /movements/:movementType/:movementId/messages/:messageId/body    v2.controllers.V2MovementsController.getMessageBody(movementType: v2.models.MovementType, movementId: v2.models.MovementId, messageId: v2.models.MessageId)
```
drnaylor-oc, line 22: If we're introducing a new endpoint, we'll need to add it to the API definition.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -228,20 +240,75 @@ class V2MovementsControllerImpl @Inject() (
...
+          json: JsObject = Json.toJson(HateoasMovementMessageResponse(movementId, messageSummary.id, messageSummary, movementType)).as[JsObject]
+          stream <- mergeStreamIntoJson(json.fields, "body", bodyStream)
+        } yield stream
+    }
+
+  private def processSmallMessage(movementId: MovementId, movementType: MovementType, messageSummary: MessageSummary, acceptHeader: String)(implicit
+    hc: HeaderCarrier
+  ): EitherT[Future, PresentationError, Source[ByteString, _]] =
+    acceptHeader match {
+      case VersionedRouting.VERSION_2_ACCEPT_HEADER_VALUE_JSON =>
```
derekspalmer, line 276: Possibly combine with process smallMessage since they have the same match conditions.

app/v2/controllers/stream/StreamingParsers.scala
```diff
@@ -80,4 +86,99 @@ trait StreamingParsers {
...
+              }
+              .intersperse(ByteString(",")) ++ END_TOKEN_SOURCE
+          )
+        )
+        .recover {
+          case NonFatal(e) => Left(PresentationError.internalServiceError(cause = Some(e)))
+        }
+    }
+
+  def jsonWrappedXmlToByteStringStream(fields: collection.Seq[(String, JsValue)]): EitherT[Future, PresentationError, Source[ByteString, _]] =
```
derekspalmer, line 151: Can we make the code within .successful(..) a function as seems duplicated in the 3 functions

app/v2/models/responses/MessageSummary.scala
```diff
@@ -35,14 +36,16 @@ object MessageSummary {
       (JsPath \ "received").read[OffsetDateTime] and
       (JsPath \ "messageType").read[MessageType] and
       (JsPath \ "body").readNullable[XmlPayload] and
-      (JsPath \ "status").readNullable[MessageStatus])(MessageSummary.apply _)
+      (JsPath \ "status").readNullable[MessageStatus] and
```
derekspalmer, line 39: Json.Format[MessageSummary] ?

app/v2/controllers/V2MovementsController.scala
```diff
@@ -228,20 +240,75 @@ class V2MovementsControllerImpl @Inject() (
     }
 
   def getMessage(movementType: MovementType, movementId: MovementId, messageId: MessageId): Action[AnyContent] =
-    (authActionNewEnrolmentOnly andThen acceptHeaderActionProvider(acceptOnlyJson = false)).async {
+    (authActionNewEnrolmentOnly andThen acceptHeaderActionProvider(jsonAndJsonWrappedXmlAcceptHeaders)).async {
+      implicit request =>
+        implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+        (for {
+          messageSummary <- persistenceService.getMessage(request.eoriNumber, movementType, movementId, messageId).asPresentation
+          acceptHeader = request.headers.get(HeaderNames.ACCEPT).get
```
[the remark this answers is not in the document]
liandra-giacomine: It shouldn't, at this stage we know there's an accept header, check acceptHeaderActionProvider

test/v2/services/ObjectStoreServiceSpec.scala
```diff
@@ -74,13 +78,39 @@ class ObjectStoreServiceSpec
...
           case Left(x)  => x
         }
     }
   }
+
+  "On getting a message from object store" - {
+    "given a successful response from the connector, should return a Right with Object Store Summary" in {
+      val movementId = MovementId("308c4a68e2cdc08f")
+      val messageId  = MessageId("123c4a68e2ele08f")
+      objectStoreService.addMessage(DownloadUrl("https://bucketName.s3.eu-west-2.amazonaws.com"), movementId, messageId)
```
sonisn1988, line 95: I think we can remove this part. not required as you are explicitly stubbed the getObject so you are always going to get something without adding anything in object store.

build.sbt
```diff
@@ -29,7 +29,8 @@ lazy val microservice = Project(appName, file("."))
       "models.Binders._",
       "java.time.OffsetDateTime",
       "v2.models.EORINumber",
-      "v2.models.MovementId"
+      "v2.models.MovementId",
+      "v2.models.MovementType"
```
sonisn1988, line 33: No need to add here. I have done it without adding ``` MovementType ``` and seems to be working.

it/v2/utils/CommonGenerators.scala
```diff
@@ -151,18 +156,6 @@ trait CommonGenerators {
     } yield MessageUpdate(status, Some(ObjectStoreURI(uri)))
   }
 
-  implicit val arbitraryObjectStoreURI: Arbitrary[ObjectStoreURI] = Arbitrary {
```
sonisn1988, line 154: This is what ObjectStoreURI should be . so should not be removed. 
Please remove above one
``` implicit lazy val arbitraryObjectStoreURI: Arbitrary[ObjectStoreURI] = Arbitrary {
    Gen.alphaNumStr.map(ObjectStoreURI(_))
  } ```

File: app/v2/controllers/stream/StreamingParsers.scala [hunk 9]
```diff
@@ -80,4 +87,99 @@ trait StreamingParsers {
       }
   }
 
+  private val START_TOKEN: Source[ByteString, NotUsed]                 = Source.single(ByteString("{"))
+  private val END_TOKEN_SOURCE: Source[ByteString, NotUsed]            = Source.single(ByteString("}"))
+  private val END_TOKEN_WITH_QUOTE_SOURCE: Source[ByteString, NotUsed] = Source.single(ByteString("\"}"))
+
+  def mergeStreamIntoJson(
```
sonisn1988, line 94: Please add test cases for this changes done in this file.

test/v2/services/ObjectStoreServiceSpec.scala
drnaylor-oc: > you are explicitly stubbed the getObject

Is that part of `ObjectStoreStub` -- does that always return something?

sonisn1988: Yes it is. if we are passing some uri that does not match path then it returns None.

drnaylor-oc: `ObjectStoreStub` isn't ours though.

If this works, then I'm happy to keep it as is.

sonisn1988: It is written by us if you can see that in fake folder and changed that with this PR only.

drnaylor-oc: No it doesn't, not the `ObjectStoreStub`, which is part of the client -- https://github.com/hmrc/object-store-client/blob/56d4e7e73aebeaa40c9284bc01bb65c7d34d1db7/object-store-client-play-28/src/main/scala/uk/gov/hmrc/objectstore/client/play/test/stub.scala#L78-L82.

You're thinking of the object-store SERVER stubbing, which uses the file system.

build.sbt
drnaylor-oc: Correct, but the code for the binder is in the wrong place (it's currently in `models.Binders`, should really be in a `v2.models.Binders`.

I was a bit confused as to why that worked.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -228,20 +240,75 @@ class V2MovementsControllerImpl @Inject() (
...
+          body <-
+            if (messageSummary.uri.isDefined) processLargeMessage(movementId, movementType, messageSummary, acceptHeader)
```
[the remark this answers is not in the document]
liandra-giacomine: In that case we would have a really big set of case statements to separate the logic for small and large so I'd rather keep the implementation in separate functions but I've made the code more readable/functional by using a match statement instead of an if statement

app/v2/controllers/V2MovementsController.scala
liandra-giacomine: as discussed, same match conditions but very different implementations so would over complicate and make the function less readable in my opinion

app/v2/controllers/stream/StreamingParsers.scala
liandra-giacomine: Sure, I'll reduce the duplication

it/v2/utils/CommonGenerators.scala
liandra-giacomine: It's still present, it's just before the arbitraryMessageSummary as it now is used in that generator

test/v2/services/ObjectStoreServiceSpec.scala
drnaylor-oc: Oh, I see what you were saying, I'd missed the part of your message saying _we'd_ written it, right.

I get what you're saying now -- https://github.com/hmrc/common-transit-convention-traders/pull/397/files/5647541c21e5a93b8ab7c2def660d355ae518062#diff-ed5f8f2141ae725f9526396be4fc46d06a963bcd0d8e7063b1ff807363f7414eR81.

Yup, I get it now -- yeah, putting the object can be ignored.

it/v2/utils/CommonGenerators.scala
drnaylor-oc: He's saying that this implementation should still be used, not simplified to `Gen.alphaNumStr.map(ObjectStoreURI(_))` (because that's not indicative of what object store URIs would look like)

test/v2/services/ObjectStoreServiceSpec.scala
liandra-giacomine: Yes it's great that you were able to just use their stub @sonisn1988 and I massively envy that. I went thought the headache of trying to replicate what happens in the stub as much as possible for in our own stub that overrides theirs so that the functionality in the put would work as expected. https://github.com/hmrc/object-store-client/blob/56d4e7e73aebeaa40c9284bc01bb65c7d34d1db7/object-store-client-play-28/src/main/scala/uk/gov/hmrc/objectstore/client/play/test/stub.scala#L96

What happens is that they have an internal object-store map to store these values, while I see your solution to [hard code](https://github.com/hmrc/transit-movements-router/blob/main/test/uk/gov/hmrc/transitmovementsrouter/fakes/objectstore/ObjectStoreStub.scala#L64) it is one way to do it, I preferred writing the code with a map such that we can first of all replicate  what happens in the stub as much as possible but second of all have flexibility. I can then ensure that my put works as expected too.

@drnaylor-oc for your understanding

drnaylor-oc: Got it, thanks.

app/v2/models/responses/MessageSummary.scala
liandra-giacomine: My bad, we do need it as we need to know which Payload type to use for the reads (JsPath \ "body").readNullable[XmlPayload]

VijayKumarTech (approved):
Testing satisfied and Happy to approve.
This PR can now be merged upon satisfactory peer review

PR #400, opened 2023-03-21
Added json content type to streamed out response

liandra-giacomine:


drnaylor-oc:
A test failed, just re-running to see if it's consistent, but we'll need to have a look at that at some point.

drnaylor-oc:
Ran fine -- it looked to be something to do with the object store stub.

PR #401, opened 2023-03-21
Check message size

liandra-giacomine:
We were previously relying on the content type header to find out the size of the payload being sent. This PR updates that check to be done by calculating the size of the payload instead. This is more reliable, as it is possible to send a content-type that doesn't match the payload size and we aren't able to get the content size if the request is streamed to us. This also resolves broke pipe issues we found in the regression tests.

I have also updated the logic in getMessage so that we check whether we have a small message stored in object-store. If so, then we should allow traders to send a JSON accept header, otherwise we return 406.


derekspalmer (approved):
LGTM

resources/public/api/conf/2.0/application.yaml
```diff
@@ -570,820 +581,805 @@ responses:
...
+        For example, this could include ‘Declaration rejected’ or Movement Reference Number (MRN) allocated messages.
+        
+        Movements older than 31 days and their related messages are archived and cannot be returned by the API.
+        
+        The response can be returned in one of the following formats based on the provided `Accept` header:
+        
+         * `application/vnd.hmrc.2.0+xml`: XML, where the message is provided without any other metadata.
+        
+         <br/><br/>
+         **Note:** You can view a transit movement only if your Economic Operators Registration and Identification (EORI) number matches the EORI number that created the transit declaration for that movement. Your EORI number is part of the bearer token that you use for API authentication.\n
```
drnaylor-oc, line 633: Remove the \n here.

drnaylor-oc:
Does this replace #400?

drnaylor-oc:
Cool. A couple more things from me on the API definition then we should be okay (properly went though it!) -- will post those in a moment.

resources/public/api/conf/2.0/application.yaml
```diff
@@ -570,820 +581,805 @@ responses:
...
+                  - description: XML payload
+                    xml:
+                      name: CustomsTransitsMovementsDeparturesMessageBodyResponse
+                      attribute: false
+                      wrapped: false
+              example:
+                >
+                n1.CC015C:
+                  @PhaseID: "NCTS5.0"
+                  ...
```
drnaylor-oc, line 668:
```yaml
              example: '<ncts:CC015C PhaseID="NCTS5.0" xmlns:ncts="http://ncts.dgtaxud.ec">...</ncts:CC015C>'
```

resources/public/api/conf/2.0/application.yaml
```diff
@@ -570,820 +581,805 @@ responses:
...
+                    xml:
+                      name: CustomsTransitsMovementsArrivalsMessageBodyResponse
+                      attribute: false
+                      wrapped: false
+              example:
+                >
                 n1.CC007C:
-                  @PhaseID: NCTS5.0
+                  @PhaseID: "NCTS5.0"
                   ...
```
drnaylor-oc, line 1352: example: '<ncts:CC007C PhaseID="NCTS5.0" xmlns:ncts="http://ncts.dgtaxud.ec">...</ncts:CC007C>'

drnaylor-oc (commented):
I also have this diff which should also be applied:

```diff
@@ -1755,13 +1747,13 @@ components:
         status: Success
         body: <ncts:CC015C PhaseID="NCTS5.0" xmlns:ncts="http://ncts.dgtaxud.ec">...</ncts:CC015C>
     CustomsTransitsMovementsDeparturesMessageBodyResponse:
-      title: CustomsTransitsMovementsDeparturesMessageBodyResponse
-      type: xml
+      title: XML payload
+      type: string
       example:
         <ncts:CC015C PhaseID="NCTS5.0" xmlns:ncts="http://ncts.dgtaxud.ec">...</ncts:CC015C>
     CustomsTransitsMovementsArrivalsMessageBodyResponse:
-      title: CustomsTransitsMovementsDeparturesMessageBodyResponse
-      type: xml
+      title: XML payload
+      type: string
       example:
         <ncts:CC007C PhaseID="NCTS5.0" xmlns:ncts="http://ncts.dgtaxud.ec">...</ncts:CC007C>
     CustomsTransitsMovementsDeparturesResponse:
```

app/v2/controllers/V2MovementsController.scala
```diff
@@ -312,7 +362,7 @@ class V2MovementsControllerImpl @Inject() (
             } else stringToByteStringStream(messageSummary.body.get.value)
```
sonisn1988, line 362: I think this will get failed in case of we call this for empty movement(empty movement is going to have any body)?

drnaylor-oc: Yup, that's true.

drnaylor-oc (approved):
I'm relatively happy with this. Some of the matchers in the tests could be a little stricter (any -> eqTo) BUT, as discussed, those tests are likely to disappear when I fold stuff into `transit-movements` in the next week or so.

drnaylor-oc:
(Only merge when tests are ready though, please)

PR #402, opened 2023-03-24
CTCP-2427: add validation on source from object store

derekspalmer:


it/v2/connectors/ValidationConnectorSpec.scala
```diff
@@ -322,6 +326,92 @@ class ValidationConnectorSpec
...
+      "then on successful validation of schema valid XML, must return NO_CONTENT" in {
+        server.stubFor(
+          post(
+            urlEqualTo("/transit-movements-validator/messages/IE015/validation")
+          )
+            .withHeader(XObjectStoreUriHeader, equalTo("resource/location"))
+            .willReturn(
+              aResponse().withStatus(NO_CONTENT)
+            )
+        )
```
sonisn1988, line 340: No need to pass extra headers. Please remove it for all down wards test cases as well

derekspalmer: These test case will fail if the response is removed!

it/v2/connectors/ValidationConnectorSpec.scala
```diff
@@ -322,6 +326,92 @@ class ValidationConnectorSpec
...
+        implicit val hc: HeaderCarrier = HeaderCarrier(extraHeaders = Seq(HeaderNames.ACCEPT -> ContentTypes.JSON))
```
drnaylor-oc, line 341: I think Sandeep meant this line -- the connector should handle ensuring the headers are correct so putting this here effectively bypasses that, and the test wouldn't catch it if this was missed.

it/v2/connectors/ValidationConnectorSpec.scala
sonisn1988: just check this part. don't think so we need to pass HeaderNames.ACCEPT -> ContentTypes.JSON? implicit val hc: HeaderCarrier = HeaderCarrier(extraHeaders = Seq(HeaderNames.ACCEPT -> ContentTypes.JSON))

it/v2/connectors/ValidationConnectorSpec.scala
derekspalmer: Yes, get it now!

sonisn1988 (approved):

derekspalmer:
Add validation using source from object-store.

PR #404, opened 2023-03-27
CTCP-2581 Message status update to success or failed

akila-sankar:


app/v2/controllers/V2MovementsController.scala
```diff
@@ -392,6 +392,24 @@ class V2MovementsControllerImpl @Inject() (
...
+            updateSmallMessageStatus(
+              request.eoriNumber,
+              movementType,
+              movementId,
+              updateMovementResponse.messageId,
+              MessageStatus.Failed
+            )
+            err
+        }
+      result = updateSmallMessageStatus(
```
liandra-giacomine, line 406: no need for the `result = ` here as the result will be implied by the  ```
_ <- routerService
         .send
``` call. Also, if you want the value of something to be evaluated in a for-comprehension composed of EitherTs then you need to use <-  rather than =

akila-sankar: updated

PR #406, opened 2023-03-30
Ctcp 2008 trader failed upload

akila-sankar:


test/v2/models/responses/TraderFailedUploadAuditRequestSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+import v2.base.TestCommonGenerators
+import v2.models.EORINumber
+import v2.models.MessageId
+import v2.models.MovementId
+
+class TraderFailedUploadAuditRequestSpec extends AnyFreeSpec with Matchers with ScalaCheckDrivenPropertyChecks with TestCommonGenerators {
+
+  private val gen = Gen.listOfN(10, Gen.alphaChar).map(_.mkString)
+
+  "when request is serialized, return an appropriate JsObject" in forAll(gen, gen, gen, arbitraryMovementType.arbitrary) {
```
derekspalmer, line 34: We have generators for these that we can use:
arbitraryMovementId
arbitraryMessageId,
arbitraryEORINumber

drnaylor-oc: or, import `org.scalacheck.Arbitrary.arbitrary` and then you can use `arbitrary[MovementId]`, `arbitrary[MessageId]` etc., which will then throw an error if the implicit generator doesn't exist.

derekspalmer (approved):

PR #407, opened 2023-03-30
Ctcp 2452 test

olakunlekuye2:


app/v2/connectors/PersistenceConnector.scala
```diff
@@ -216,4 +228,23 @@ class PersistenceConnectorImpl @Inject() (httpClientV2: HttpClientV2, appConfig:
...
+    messageId: MessageId,
+    source: Source[ByteString, _]
+  )(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): Future[Unit] = {
+    val url = appConfig.movementsUrl.withPath(postMessageBodyUrl(eoriNumber, movementType, movementId, messageId))
+
+    httpClientV2
+      .post(url"$url")
```
sonisn1988, line 244: Here we need to pass the message Type as part of header. Just look on how it is passed for other endpoint in this class.Add X-Message-Type with header

app/v2/controllers/V2MovementsController.scala
```diff
@@ -582,6 +612,51 @@ class V2MovementsControllerImpl @Inject() (
...
+          _ <- updateMovementResponse(messageUpdate(MessageStatus.Processing)).asPresentation
+          _ <- validationService
+            .validateXml(messageType, source)
+            .asPresentation
+            .leftMap {
+              err =>
+                updateMovementResponse(messageUpdate(MessageStatus.Failed)).value
+                err
+            }
+          _ <- persistenceService.updateMessageBody(eori, movementType, movementId, messageId, source).asPresentation
```
sonisn1988, line 639: @drnaylor-oc  With this newly created endpoint to update the body I can see we are checking the status on transit-movements code and if that status is Pending then only it is going to update the body otherwise it will return error
https://github.com/hmrc/transit-movements/blob/main/app/uk/gov/hmrc/transitmovements/controllers/MessageBodyController.scala#L115
Now here at this point of time we will having message status as Processing hence will be getting error from transit-movements.  So either Api or Transit-movements we need to modify the code. Please suggest.

app/v2/connectors/PersistenceConnector.scala
olakunlekuye2: Done

app/v2/controllers/V2MovementsController.scala
drnaylor-oc: The only time that a body should not exist is if the status is `Pending`. It should be immutable after that. So, if something is setting the status to `Processing` before uploading the body, then that's wrong -- hence, the API needs to change.

app/v2/connectors/PushNotificationsConnector.scala
```diff
@@ -24,11 +24,15 @@ import metrics.HasMetrics
 import metrics.MetricsKeys
 import play.api.http.HeaderNames
 import play.api.http.MimeTypes
+import play.api.http.Status.ACCEPTED
 import play.api.http.Status.NO_CONTENT
+import play.api.libs.json.JsValue
 import play.api.libs.json.Json
+//import play.libs.Json
```
sonisn1988, line 31: remove this commented line.

test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -4369,32 +4418,114 @@ class V2MovementsControllerSpec
 
                       verify(mockXmlParsingService, times(1)).extractMessageType(any(), any())(any(), any())
 
-                      verify(mockValidationService, times(1)).validateLargeMessage(any(), any())(any(), any())
+//                      verify(mockValidationService, times(1)).validateLargeMessage(any(), any())(any(), any())
```
sonisn1988, line 4421: remove commented line

test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -4818,7 +5017,7 @@ class V2MovementsControllerSpec
 
                 verify(mockXmlParsingService, times(1)).extractMessageType(any(), any())(any(), any())
 
-                verify(mockValidationService, times(1)).validateLargeMessage(eqTo(MessageType.DeclarationAmendment), any())(any(), any())
+//                verify(mockValidationService, times(1)).validateLargeMessage(eqTo(MessageType.DeclarationAmendment), any())(any(), any())
```
sonisn1988, line 5020: remove commented line

sonisn1988 (changes requested):

app/v2/controllers/V2MovementsController.scala
sonisn1988: @olakunlekuye2  in that case you are going to set message status as Processing post this call. Currently you are setting it before making to this call. Just shuffle that code and then I hope it should be fine.

drnaylor-oc: You don't need to set the status at all, that call will do it for you.

test/v2/controllers/V2MovementsControllerSpec.scala
olakunlekuye2: Done

test/v2/controllers/V2MovementsControllerSpec.scala
olakunlekuye2: Done

app/v2/connectors/PushNotificationsConnector.scala
olakunlekuye2: Done

sonisn1988: still there see above.

app/v2/controllers/V2MovementsController.scala
olakunlekuye2: Done

sonisn1988 (changes requested):
Please resolve the merge conflict

app/v2/connectors/PushNotificationsConnector.scala
sonisn1988: remove this part //import play.libs.Json

app/v2/controllers/V2MovementsController.scala
```diff
@@ -475,68 +470,103 @@ class V2MovementsControllerImpl @Inject() (
...
+                      _ <- persist(messageUpdate(MessageStatus.Processing)).asPresentation
+                      sendMessage <- routerService
+                        .sendLargeMessage(
+                          messageType,
+                          eori,
+                          movementId,
+                          messageId,
+                          ObjectStoreURI(objectSummary.location.asUri)
+                        )
+                        .asPresentation
```
sonisn1988, line 547: by Mistake Derek has removed below code from here. So please add that back again here

 ``` .leftMap {
  err =>
    persist(messageUpdate(MessageStatus.Failed)).value
    err
} ```

drnaylor-oc:
Can we re-play the updates on https://github.com/hmrc/common-transit-convention-traders/pull/412 instead please? I made some significant changes to that branch, both to reduce global state and also to enable some tests that weren't enabled before because they'd been specified incorrectly...

PR #411, opened 2023-04-12
Ctcp 2430

derekspalmer:


app/v2/connectors/AuditingConnector.scala
```diff
@@ -50,14 +53,17 @@ class AuditingConnectorImpl @Inject() (httpClient: HttpClientV2, appConfig: AppC
...
+    ec: ExecutionContext
+  ): Future[Unit] =
     withMetricsTimerAsync(MetricsKeys.AuditingBackend.Post) {
       _ =>
         val url = appConfig.auditingUrl.withPath(auditingRoute(auditType))
 
         httpClient
           .post(url"$url")
-          .transform(_.addHttpHeaders(HeaderNames.CONTENT_TYPE -> contentType))
+          .transform(_.addHttpHeaders(HeaderNames.CONTENT_TYPE -> contentType, "ContentLength" -> contentLength.toString))
```
drnaylor-oc, line 66: If we're going to use a header for the length, as I said in the other PR, it needs to start with `X-`

derekspalmer: Yep.

derekspalmer: Yep.

prashanthmamidi (approved):

app/v2/services/AuditingService.scala
```diff
@@ -33,15 +33,21 @@ import scala.util.control.NonFatal
 @ImplementedBy(classOf[AuditingServiceImpl])
 trait AuditingService {
 
-  def audit(auditType: AuditType, source: Source[ByteString, _], contentType: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Unit]
+  def audit(auditType: AuditType, source: Source[ByteString, _], contentType: String, contentLength: Long = 0)(implicit
```
drnaylor-oc, line 36: I know this has been merged and all... but what does a `contentLength` of 0 mean? This sends the header with a content length of 0... and what does the auditing service do with that?

I checked -- and it basically assumes that the message is small. Is that really what we should be doing? Yes, it does that in the absence of a header too, but at least there the intent is crystal clear. Sending a `X-ContentSize` header of `0` for a non-zero body isn't right.

This is something that really should be an `Option` and then not send the header if we can't supply the size. We're now relying on a magic number to determine behaviour, rather than the existence or absence of a header. 

This needs rectifying because this isn't a great thing to do.

PR #413, opened 2023-04-14
[CTCP-2813] check if the movementId exists for the movementType before adding a message

prashanthmamidi:
Added the check to ensure the movementId exists for the given movementType before adding a message.


app/v2/controllers/V2MovementsController.scala
```diff
@@ -426,6 +426,7 @@ class V2MovementsControllerImpl @Inject() (
         request.body.runWith(Sink.ignore)
 
         (for {
+          _                      <- persistenceService.getMovement(request.eoriNumber, movementType, movementId).asPresentation
```
sonisn1988, line 429: I think this should be handled in transit-movements not here.

prashanthmamidi: I think it's better we leave this here to keep this simple and avoid business logic in the db layer.  And also attachLargeMessage in transit-movements we do not send movementType

sonisn1988: movementId is unique in database so when we are making call to db from api side then in transit-movements we should first check if the movementId exist if so then good otherwise 404 from there to api side. we can manage this on transit-movements without Knowing the movementType.

sonisn1988 (approved):

app/v2/controllers/V2MovementsController.scala
drnaylor-oc: Ideally, yes, such a check should happen in `transit-movements`. Given how our routes are set up, however, we do not pass the movement type to the backend so what Prashanth is saying is totally correct.

I'm minded to accept this, but add a low priority tech debt ticket to address this later.

sonisn1988 (approved):

PR #414, opened 2023-04-14
CTCP-2452: Rewrite and simplify attaching a message via Upscan

drnaylor-oc:
We were getting a bit messy with attaching a file from Upscan and we've gotten ourselves a little confused because of the original direction of object-store being used everywhere. This PR takes things back to the basic principles. A couple of things to note:

* Platops wanting us to rationalise the object store code is doing us a favour here, as we now call transit movements with a stream and _that's_ what cares about storage
* We still have the router wanting an object-store URI. Instead, we should just pass the stream and let the router decide what goes where. That'll come up in another PR later. We'll make the router return 200 for a submission success, and 202 for submission still pending (SDES). We can use that to then determine if we push a notification. For now, we just use the isLarge check (but in the future, the router might have a different size limit to the API)


derekspalmer (approved):
Looks ok.
It should be a lot tidier once functionality is moved out and into the router and transit-movements.

PR #415, opened 2023-04-18
Send push notification when submission has completed 

olakunlekuye2:


drnaylor-oc:
You should add `verify` calls to the various tests that call those methods to make sure the push notifications are actually being called.

In particular, you should verify that there are NO calls to PPNS in the case of a successful large message.

sonisn1988 (commented):
can you please rebase this with main as some of the changes has been already merged with main but still can see with this PR as well

app/v2/controllers/V2MovementsController.scala
```diff
@@ -533,7 +533,8 @@ class V2MovementsControllerImpl @Inject() (
                  |Reason: ${failureDetails.failureReason}
                  |Message: ${failureDetails.message}""".stripMargin)
 
-            // TODO: Push notification on failure
+            pushNotificationsService
+              .postPpnsNotification(movementId, messageId, Json.toJson(HateoasMovementUpdateResponse(movementId, messageId, movementType, None)))
```
sonisn1988, line 537: @drnaylor-oc  I think we need to send upscan failure details here  UpscanFailedResponse(reference, failureDetails) instead of HateoasMovementUpdateResponse(movementId, messageId, movementType, None)?

drnaylor-oc: Yes, a failure shouldn't be returning a Hateoas object. This should just be a bad request presentation error.

sonisn1988: @olakunlekuye2  Please change this as suggested.

olakunlekuye2: Done

sonisn1988 (approved):

app/v2/controllers/V2MovementsController.scala
```diff
@@ -533,7 +533,8 @@ class V2MovementsControllerImpl @Inject() (
                  |Reason: ${failureDetails.failureReason}
                  |Message: ${failureDetails.message}""".stripMargin)
 
-            // TODO: Push notification on failure
```
sonisn1988, line 536: @drnaylor-oc  I think we also need to set the status as failed here in db for this use case?

app/v2/controllers/V2MovementsController.scala
```diff
@@ -533,7 +533,8 @@ class V2MovementsControllerImpl @Inject() (
...
+            pushNotificationsService
+              .postPpnsNotification(movementId, messageId, Json.toJson(UpscanFailedResponse(reference, failureDetails)))
```
drnaylor-oc, line 537: What does `UpscanFailedResponse` look like?

drnaylor-oc: Wait, no, this isn't appropriate to send.

This should really be a `PresentationError.badResponse` with a message that's appropriate, perhaps something like "Uploaded file not accepted". We should NOT be sending back potentially internal details.

drnaylor-oc (changes requested):
The push notification for an upscan failure should be a `BAD_REQUEST`, not just replaying the upscan failure (which is an internal thing(

sonisn1988 (dismissed):

drnaylor-oc (dismissed):

sonisn1988 (approved):

PR #420, opened 2023-05-03
Push a SUCCESS notification rather than Hateoas response when a small…

sonisn1988:
… message is sent via the large message route (common-transit-convention-traders)


drnaylor-oc (approved):
We'll merge this end of day when testing is done for the day.

PR #432, opened 2023-05-31
Incorrect status code returned when filtering returns no messages/mov…

sonisn1988:
…ements

### DIFF
--- a/app/v2/services/PersistenceService.scala
+++ b/app/v2/services/PersistenceService.scala
@@ -161,8 +161,7 @@ class PersistenceServiceImpl @Inject() (persistenceConnector: PersistenceConnect
         .getMessages(eori, movementType, movementId, receivedSince)
         .map(Right(_))
         .recover {
-          case UpstreamErrorResponse(_, NOT_FOUND, _, _) => Left(PersistenceError.MovementNotFound(movementId, movementType))
-          case NonFatal(thr)                             => Left(PersistenceError.UnexpectedError(Some(thr)))
+          case NonFatal(thr) => Left(PersistenceError.UnexpectedError(Some(thr)))
         }
     )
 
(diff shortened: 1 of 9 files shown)


resources/public/api/conf/2.0/application.yaml
```diff
@@ -1048,19 +1035,6 @@ paths:
...
-            application/json:
-              schema:
-                allOf:
-                  - $ref: '#/components/schemas/NotFound'
-                  - example:
-                      code: NOT_FOUND
-                      message: Supplied arrival not found or does not exist or has been archived or is not available to the EORI number.
-              example:
-                code: NOT_FOUND
-                message: Supplied arrival not found or does not exist or has been archived or is not available to the EORI number.
```
drnaylor-oc, line 1063: Same here

app/v2/services/PersistenceService.scala
```diff
@@ -161,8 +161,7 @@ class PersistenceServiceImpl @Inject() (persistenceConnector: PersistenceConnect
         .getMessages(eori, movementType, movementId, receivedSince)
         .map(Right(_))
         .recover {
-          case UpstreamErrorResponse(_, NOT_FOUND, _, _) => Left(PersistenceError.MovementNotFound(movementId, movementType))
```
drnaylor-oc, line 164: NOT_FOUND could still happen here with a invalid movement ID.

test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -3360,16 +3361,15 @@ class V2MovementsControllerSpec
...
           val request = FakeRequest("GET", "/", FakeHeaders(), Source.empty[ByteString])
           val result  = sut.getMessageIds(movementType, movementId, None)(request)
 
-          status(result) mustBe NOT_FOUND
-          contentAsJson(result) mustBe Json.obj(
-            "code"    -> "NOT_FOUND",
-            "message" -> s"${movementType.movementType.capitalize} movement with ID ${movementId.value} was not found"
+          status(result) mustBe OK
+          contentAsJson(result) mustBe Json.toJson(
+            HateoasMovementMessageIdsResponse(movementId, List.empty[MessageSummary], None, movementType)
```
drnaylor-oc, line 3372: When no movement is found for the list messages endpoint, this should still return `NOT_FOUND` -- because the _movement_ was not found,

drnaylor-oc (dismissed):
My previous comments made me realise you'd gone a bit too far.

* On **get movements**, no movements returned is an empty list + 200
* On **get messages for a movement**:
  * If the **movement ID is incorrect** return 404
  * If the **movement ID is correct** but no messages are found, return empty list + 200

drnaylor-oc (approved):
Await regression tests and for the LRN stuff to be merged first before merging, but looking at this I see no real issue.

I would like sight of the regression tests before this merges too, just to make sure that there isn't any confusion there.

drnaylor-oc (approved):
As before, please await regression tests (and I would like quick sight of those before we merge)

PR #437, opened 2023-06-13
Updated ErrorCode

olakunlekuye2:


drnaylor-oc:
This isn't the right change to make.

From the requirements, the final bullet point is

> Still return 400 to trader, this is an internal change

What CTCP-3110 is all about is making everything consistent internally. The _validator_ should return 200 because the validator is expecting some requests to not fit the validation rules and it returns a validation object -- it's not a bad request there because its purpose is to check these things. 

However, overall, the trader should still be given a HTTP 400, and so this bit is okay, this change would mean that traders would get a 200 for an invalid payload.

I realise that might seem confusing, but the basic principle is:

* `transit-movements-validator` should return `200` if a payload isn't valid, because the validator _expects_ a response that might be invalid. A `400` would be returned from the validator if it can't _attempt_ to validate (given an message type that doesn't exist, for example)
* In the wider context, `common-transit-convention-traders` should return `400`, because although it's the validator's job to validate, overall, the trader provided something invalid that the wider service did NOT expect.

This change will cause traders to get a 200 on a business validation error. That's not correct, so this PR should be closed. However, in `common-transit-convention-traders`, the validation service/connector DO need to be looked at, to make sure that when a 200 is returned from the validator, both schema and business rule validation messages can be handled.

Does that make sense?

drnaylor-oc:
(also, this isn't your fault but the fact that this change didn't break any tests isn't great...)

drnaylor-oc:
It seems six tests failed, they'll need to be corrected.

it/v2/connectors/ValidationConnectorSpec.scala
```diff
@@ -133,13 +122,7 @@ class ValidationConnectorSpec
...
               aResponse()
                 .withStatus(BAD_REQUEST)
-                .withBody(
-                  Json.stringify(
-                    Json.toJson(
-                      PresentationError.badRequestError("Invalid message type") // The message doesn't matter here
-                    )
-                  )
-                )
+                .withBody("Not a JSON string") // replace with invalid JSON string
```
derekspalmer, line 125: Comment?

derekspalmer (approved):
Looks ok - can we address the comments or remove them if they're redundant?

it/v2/connectors/ValidationConnectorSpec.scala
olakunlekuye2: probably tech debt

PR #441, opened 2023-06-16
Updated the V2MovementsController

olakunlekuye2:


app/v2/controllers/V2MovementsController.scala
```diff
@@ -434,28 +434,59 @@ class V2MovementsControllerImpl @Inject() (
...
-                    movementType,
-                    updatedSince,
-                    movementEORI,
-                    movementReferenceNumber,
-                    page,
-                    count,
-                    receivedUntil,
-                    localReferenceNumber
-                  )
+        (page, count) match {
```
drnaylor-oc, line 437: We'd need to do this for the `getMessages` endpoint as well, this solution isn't very reusable.

Instead, why don't we make a method:

```scala
def ensurePositive(parameter: Option[Int], keyName: String): EitherT[Future, PresentationError, Unit] = 
    parameter match {
        case Some(x) if x <= 0 => EitherT.leftT(BadRequest(...))
        case _ => EitherT.rightT((): Unit)
    }
``` 

and then we can just use this function in the for comprehension:

```scala
for {
    _ <- ensurePositive(page, "page")
    _ <- ensurePositive(page, "count")
   ....
} yield ....
```

and that way everything is reusable?

olakunlekuye2: Okay I'll do that

sonisn1988 (dismissed):
LGTM

app/v2/controllers/V2MovementsController.scala
```diff
@@ -398,13 +390,24 @@ class V2MovementsControllerImpl @Inject() (
...
-          .asPresentation
-          .fold(
-            presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
-            response => Ok(Json.toJson(HateoasMovementMessageIdsResponse(movementId, response, receivedSince, movementType, page, count, receivedUntil)))
-          )
+        val pageValidation: EitherT[Future, PresentationError, Unit] =
+          ensurePositive(page.map(_.value), "page")
+
+        val countValidation: EitherT[Future, PresentationError, Unit] =
+          ensurePositive(count.map(_.value), "count")
```
drnaylor-oc, line 397: Don't make these variables, use the directly in the for comprehension. If you do that, if the first fails, the second won't execute. No point wasting cycles.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -431,31 +443,53 @@ class V2MovementsControllerImpl @Inject() (
...
-                    count,
-                    receivedUntil,
-                    localReferenceNumber
-                  )
+
+        val pageValidation: EitherT[Future, PresentationError, Unit] =
+          ensurePositive(page.map(_.value), "page")
+
+        val countValidation: EitherT[Future, PresentationError, Unit] =
+          ensurePositive(count.map(_.value), "count")
```
drnaylor-oc, line 454: Same here.

app/v2/controllers/V2MovementsController.scala
sonisn1988: I was thinking the same. Also I think if he is going to change then **result Val** also can be removed and can be Folded just like earlier it was done.

sonisn1988 (approved):

PR #445, opened 2023-06-21
CTCP-3301 restrict duplicate lrn from EIS

akila-sankar:


sonisn1988 (commented):
Please cover one scenario to return this duplicate error in ``` V2MovementsControllerSpec``` to make sure all are covered.

sonisn1988 (approved):
LGTM

drnaylor-oc:
This looks okay. Because we know what the router error will look like, and that this doesn't otherwise affect anything, this is good to merge.

(For the record, the one thing I did look at is how you parsed the message from the router -- and you've returned an `lrn` field, which is great, so I've no complaints here).

PR #450, opened 2023-06-26
Ctcp 3346 page not found

derekspalmer:


app/v2/services/PersistenceService.scala
```diff
@@ -226,12 +229,34 @@ class PersistenceServiceImpl @Inject() (persistenceConnector: PersistenceConnect
...
       .recover {
         case NonFatal(thr) => Left(PersistenceError.UnexpectedError(Some(thr)))
       }
   }
 
+  private def isEmptyPage(page: Option[PageNumber], numberOfItemsForPage: Int): Boolean =
+    (page, numberOfItemsForPage) match {
+      case (Some(PageNumber(1)), _)                                     => false
+      case (Some(_), numberOfItemsForPage) if numberOfItemsForPage == 0 => true
+      case (None, _)                                                    => true
```
drnaylor-oc, line 244: Isn't `None` the same as page 1 -- in which case, this should also be false?

test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -3364,8 +3364,11 @@ class V2MovementsControllerSpec
           }
       }
 
-      "when no movement is found" in forAll(arbitraryMovementId.arbitrary, arbitraryPageNumber.arbitrary, arbitraryItemCount.arbitrary) {
-        (movementId, pageNumber, itemCount) =>
+      // TODO: should this be MovementNotFound or PageNotFound ?
```
drnaylor-oc, line 3367: This is the message list one, right?

* If the **movement** is not found, this is `MovementNotFound`. It forms part of the path, so will return 404.
* If the movement is found, but there are no messages, that's `PageNotFound`.

app/v2/services/PersistenceService.scala
derekspalmer: Not sure, but if that's the case will change.

test/v2/controllers/V2MovementsControllerSpec.scala
derekspalmer: Yep, agree - my misunderstanding.

drnaylor-oc (commented):
I looked at the tests to ensure we're doing the right thing -- just one test here caught my eye.

drnaylor-oc:
Another test if we don't have it for get list of messages -- the error message from an invalid movement has higher priority over a page number > 2 -- so perhaps that's a test we should add too, that is:

If page two of a list of messages is requested for an movement that doesn't exist, then a message indicating the movement is invalid should be displayed

drnaylor-oc (approved):
I'm happy if QA is happy.

PR #453, opened 2023-06-27
Remove redundancy from v2 OAS, fix examples, fix schema

drnaylor-oc:
The automated conversion from RAML to OAS left the OpenAPI specs in a bit of a weird state, but functional for our needs. However, this has caused us to miss a few things when updating the spec, and has caused duplication that didn't need to be there.

This clears that up. This also fixes a couple of examples to use multi-examples.

### DIFF
--- a/resources/public/api/conf/2.0/application.yaml
+++ b/resources/public/api/conf/2.0/application.yaml
@@ -216,15 +216,9 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/NotFound'
-              examples:
-                MovementNotFound:
-                  value:
-                    code: NOT_FOUND
-                    message: Supplied departure not found or does not exist or has been archived or is not available to the EORI number.
-                PageNotFound:
-                  value:
-                    code: NOT_FOUND
-                    message: The requested page does not exist.
+              example:
+                code: NOT_FOUND
+                message: The requested page does not exist
         '406':
           description: Not Acceptable
           content:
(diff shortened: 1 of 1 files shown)


derekspalmer (approved):
Should we have a message along the lines of:
s"${movementType.movementType.capitalize} movement with ID ${movementId.value} was not found" ? for /messages?

fmcgrath (dismissed):
Superficially, looks OK. Hopefully, the changes mesh with changes that Derek made to the definition for NOT FOUND.

drnaylor-oc:
> Should we have a message along the lines of:
> s"${movementType.movementType.capitalize} movement with ID ${movementId.value} was not found" ? for /messages?

Actually, yes, I do need to change a couple of examples to that as that's what we have now. I'll do it tomorrow.

derekspalmer (commented):
I did try using the 'examples:' field but couldn't get it to render.
Where did you find the layout or was it from following the schema?

drnaylor-oc:
https://swagger.io/docs/specification/adding-examples/ -- search for the text "Multiple examples in response bodies".

resources/public/api/conf/2.0/application.yaml
```diff
@@ -1104,16 +947,10 @@ paths:
...
-                  - $ref: '#/components/schemas/NotFound'
-                  - example:
-                      code: NOT_FOUND
-                      message: Supplied arrival not found or does not exist or has been archived or is not available to the EORI number.
-                               Not found if the page requested is greater than the total number of pages.
+                $ref: '#/components/schemas/NotFound'
               example:
                 code: NOT_FOUND
-                message: Supplied arrival not found or does not exist or has been archived or is not available to the EORI number.
-                         Not found if the page requested is greater than the total number of pages.
```
drnaylor-oc, line 1116: There is no pagination on this endpoint (GET arrivals/{arrivalID}) so I removed this part from the examples.

derekspalmer (approved):

drnaylor-oc:
Did a final check, loaded it up in the api documention frontend viewer and it all parsed and all looked good. Merging now.

PR #454, opened 2023-06-28
CTCP-3376: Prevent use of negative years in query strings

drnaylor-oc:


prashanthmamidi (approved):

drnaylor-oc:
Before I merge this, I'm going to check v1/P4 and run tests there, because this potentially affects that too.

PR #457, opened 2023-07-03
[CTCP-3155] updated the logic to fetch the messageType from the rootNode

prashanthmamidi:


app/v2/services/JsonMessageParsingService.scala
```diff
@@ -66,10 +72,12 @@ class JsonMessageParsingServiceImpl @Inject() (implicit materializer: Materializ
...
-            messageTypeList.find(_.rootNode == root) match {
+            val jsonNode: JsonNode      = mapper.readTree(mt)
+            val rootNode                = jsonNode.fields().next().getKey
+            val messageTypeFromRootNode = rootNode.split(":")(1)
+            messageTypeList.find(_.rootNode == messageTypeFromRootNode) match {
               case Some(messageType) => Right(messageType)
-              case None              => Left(MessageTypeNotFound(root))
+              case None              => Left(MessageTypeNotFound(messageTypeFromRootNode))
             }
         }
```
drnaylor-oc, line 82: This is just going to load the whole thing into memory which we really want to avoid -- if we stick a 5MB file in here this could get big in memory as it parses and stores it in memory. It defeats the whole point of streaming.

Can we use the Jackson Streaming API here and find the first `JsonToken.FIELD_NAME` -- something like https://stackoverflow.com/a/13466740 (though not exactly, this is Java like...), then close the stream after?

You may need to convert the stream into an `InputStream` using one of the `StreamConverters` to do this.

prashanthmamidi: yes I  have tried the below  earlier `    EitherT(
      Using(source.runWith(StreamConverters.asInputStream(20.seconds))) {
        jsonInput =>
          val jsonNode: JsonNode      = mapper.readTree(jsonInput)
          val rootNode                = jsonNode.fields().next().getKey
          val messageTypeFromRootNode = rootNode.split(":")(1)
          messageTypeList.find(_.rootNode == messageTypeFromRootNode) match {
            case Some(messageType) => Right(messageType)
            case None              => Left(MessageTypeNotFound(messageTypeFromRootNode))
          }
      }.recover {
        case _ => Left(ExtractionError.MalformedInput)
      }
    )`

I have run into cats lib conversion problems from Try to Future  - play.sbt.PlayExceptions$CompilationException: Compilation error[type mismatch;
  scala.util.Try[scala.util.Either[scala.Product with v2.models.errors.ExtractionError with java.io.Serializable, v2.models.request.MessageType]]|scala.concurrent.Future[scala.util.Either[v2.models.errors.ExtractionError, v2.models.request.MessageType]]]

drnaylor-oc: Wrap the `Using` statement in `Future.fromTry(...)`

prashanthmamidi: thanks, updated

derekspalmer (approved):
LGTM.

PR #458, opened 2023-07-04
CTCP-3475: Add client ID to Upscan callback to enable it to pass through to the stub

drnaylor-oc:
On External Test, a message sent via Upscan is not getting through to ERMIS as the Upscan callback does not have the client ID. This corrects that by:

* Appending the client ID to the callback URL in the initiate call
* Reading the client ID from the callback method and adding it to the `HeaderCarrier`

Some tests have been modified to test this.

### DIFF
--- a/app/v2/controllers/V2MovementsController.scala
+++ b/app/v2/controllers/V2MovementsController.scala
@@ -38,6 +38,7 @@ import play.api.libs.json.JsObject
 import play.api.libs.json.Json
 import play.api.mvc._
 import routing.VersionedRouting
+import uk.gov.hmrc.http
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.http.HeaderCarrierConverter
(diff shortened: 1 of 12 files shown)


app/v2/controllers/V2MovementsController.scala
```diff
@@ -605,10 +613,28 @@ class V2MovementsControllerImpl @Inject() (
...
+        val originalHc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
+
+        // If the client ID is provided, ensure we add it to the headers.
+        val clientIdHeader = clientId
+          .map[Seq[(String, String)]](
+            id => Seq(XClientIdHeader -> id.value)
+          )
+          .getOrElse(Seq.empty)
+
+        // We add it to otherHeaders as hc.headers(...) doesn't check extraHeaders, and mirrors what would happen via
```
derekspalmer, line 634: :+1:

derekspalmer (approved):

PR #462, opened 2023-07-31
Submitted Json messages less than or equals to 5MB should be rejected…

sonisn1988:
… if the converted XML is over 5MB(common-transit-convention-traders)

### DIFF
--- a/app/v2/utils/StreamWithFile.scala
+++ b/app/v2/utils/StreamWithFile.scala
@@ -27,10 +27,12 @@ import play.api.Logging
 import play.api.libs.Files.TemporaryFileCreator
 import v2.models.errors.PresentationError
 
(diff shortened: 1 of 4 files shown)


app/v2/utils/StreamWithFile.scala
```diff
@@ -39,11 +41,21 @@ trait StreamWithFile {
     src: Source[ByteString, _]
```
prashanthmamidi, line 41: do we need size for withReusableSource() as well? if not better not to change this method

sonisn1988: Here we are not using size is you can see just returning source. This part is updated to minimize the code delicacy.

prashanthmamidi (approved):

PR #465, opened 2023-08-16
Force maximum limit of 5000 messages/movements for all GET list API's…

sonisn1988:
… (common-transit-convention-traders)

### DIFF
--- a/app/v2/controllers/V2MovementsController.scala
+++ b/app/v2/controllers/V2MovementsController.scala
@@ -468,11 +468,11 @@ class V2MovementsControllerImpl @Inject() (
         EitherT.rightT(())
     }
 
-  def determineMaxPerPageCount(parameter: Option[Long]): Option[ItemCount] =
+  def determineMaxPerPageCount(parameter: Option[Long]): ItemCount =
     parameter match {
-      case Some(x) if x <= config.maxItemsPerPage => Some(ItemCount(x))
-      case Some(x) if x > config.maxItemsPerPage  => Some(ItemCount(config.maxItemsPerPage))
-      case None                                   => None
+      case Some(x) if x <= config.maxItemsPerPage => ItemCount(x)
+      case Some(x) if x > config.maxItemsPerPage  => ItemCount(config.maxItemsPerPage)
+      case None                                   => ItemCount(config.defaultItemsPerPage)
     }
 
   def getMovements(
--- a/app/v2/controllers/V2MovementsController.scala
+++ b/app/v2/controllers/V2MovementsController.scala
@@ -435,8 +435,9 @@ class V2MovementsControllerImpl @Inject() (
         (for {
           _ <- ensurePositive(page.map(_.value), "page")
           _ <- ensurePositive(count.map(_.value), "count")
+          perPageCount = determineMaxPerPageCount(count.map(_.value))
           response <- persistenceService
-            .getMessages(request.eoriNumber, movementType, movementId, receivedSince, page, count, receivedUntil)
+            .getMessages(request.eoriNumber, movementType, movementId, receivedSince, page, perPageCount, receivedUntil)
             .asPresentation
         } yield response).fold(
           presentationError => Status(presentationError.code.statusCode)(Json.toJson(presentationError)),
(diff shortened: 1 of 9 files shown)


app/v2/controllers/V2MovementsController.scala
```diff
@@ -467,6 +468,13 @@ class V2MovementsControllerImpl @Inject() (
         EitherT.rightT(())
     }
 
+  def determineMaxPerPageCount(parameter: Option[Long]): Option[ItemCount] =
+    parameter match {
+      case Some(x) if x <= config.maxItemsPerPage => Some(ItemCount(x))
+      case Some(x) if x > config.maxItemsPerPage  => Some(ItemCount(config.maxItemsPerPage))
+      case None                                   => None
```
drnaylor-oc, line 475: This should also be limited to the max items per page.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -484,6 +492,7 @@ class V2MovementsControllerImpl @Inject() (
         (for {
           _ <- ensurePositive(page.map(_.value), "page")
           _ <- ensurePositive(count.map(_.value), "count")
+          perPageCount = determineMaxPerPageCount(count.map(_.value))
```
drnaylor-oc, line 495: Should also be done for messages.

resources/public/api/conf/2.0/application.yaml
```diff
@@ -139,7 +139,7 @@ paths:
 
         You can filter for departure declarations updated before and/or after a specific date and time, and optionally filter for departure declarations with a specific EORI.
 
-        For a multi-page query you can filter for a specific page, and optionally the number of items per page.
+        For a multi-page query you can filter for a specific page, and optionally the number of items per page.Maximum per page count is configured as 5000.
```
drnaylor-oc, line 142: I wouldn't bother putting this in.

app/v2/controllers/V2MovementsController.scala
sonisn1988: In that case do we need to remove ```defaultItemsPerPage```?  As there will not be any use of ```defaultItemsPerPage``` if I'll change this to ```maxItemsPerPage```

app/v2/controllers/V2MovementsController.scala
sonisn1988: This is done for messages at line number 438.

app/v2/controllers/V2MovementsController.scala
drnaylor-oc: Ah, right, yes, there's a default that gets set.

In that case, I'd like to see this logic modified because it's confusing to have this in two places like this. Choosing the page count should be done in one place, so we should move the default page count logic to here and have this just return an `Int`.

drnaylor-oc: Well `ItemCount`.

test/v2/services/PersistenceServiceSpec.scala
```diff
@@ -195,7 +188,7 @@ class PersistenceServiceSpec
       summaries =>
         val expected = PaginationMessageSummary(TotalCount(summaries.length), summaries)
         when(
-          mockConnector.getMessages(EORINumber(any()), any(), MovementId(any()), any(), eqTo(Some(PageNumber(2))), eqTo(Some(ItemCount(30))), any())(
+          mockConnector.getMessages(EORINumber(any()), any(), MovementId(any()), any(), eqTo(Some(PageNumber(2))), ItemCount(eqTo(ItemCount(30).value)), any())(
```
drnaylor-oc, line 191: `ItemCount(eqTo(ItemCount(30).value))` is really redundant -- just needs to be `ItemCount(eqTo(30))` (and the same elsewhere in the tests)

sonisn1988: I already tried the way you are suggesting but I was getting ```NullPointerException```.Just now also I checked and I am getting ```NullPointerException``` when I am using ```ItemCount(eqTo(30))```

app/v2/controllers/V2MovementsController.scala
```diff
@@ -467,6 +468,13 @@ class V2MovementsControllerImpl @Inject() (
         EitherT.rightT(())
     }
 
+  def determineMaxPerPageCount(parameter: Option[Long]): ItemCount =
+    parameter match {
+      case Some(x) if x <= config.maxItemsPerPage => ItemCount(x)
+      case Some(x) if x > config.maxItemsPerPage  => ItemCount(config.maxItemsPerPage)
+      case None                                   => ItemCount(config.defaultItemsPerPage)
+    }
```
drnaylor-oc, line 476: Kind of straying into optimisation territory here, but I don't really like this pattern match because it's doing a type match when it doesn't need to (plus one of the if guards is useless).

I wonder if we'd be better served by calling

```scala
parameter
    .map(count => Math.min(config.maxItemsPerPage, count))
   .getOrElse(ItemCount(config.defaultItemsPerPage))
```

That way we avoid reflection in the match and the runtime can make better decisions about how to make the comparison.

drnaylor-oc (approved):

PR #469, opened 2023-08-29
Remove logger config entries from configuration

drnaylor-oc:
They should be overridden by environment specific configuration anyway.

Raising for visibility -- testing this in dev against P4 code.


drnaylor-oc:
It kind of worked, but only locally, I couldn't get the logger to log things on development. I don't really know what's going on if it's not responding to the `uk.gov` logger -- perhaps worth another look at some point but for now it doesn't matter, we've worked out what we needed to work out anyway.

PR #472, opened 2023-09-19
CTCP-3908: make accept header checks case-insensitive

derekspalmer:


app/v2/controllers/V2MovementsController.scala
```diff
@@ -280,7 +280,7 @@ class V2MovementsControllerImpl @Inject() (
         implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
         (for {
           messageSummary <- persistenceService.getMessage(request.eoriNumber, movementType, movementId, messageId).asPresentation
-          acceptHeader = request.headers.get(HeaderNames.ACCEPT).get
+          acceptHeader = request.headers.get(HeaderNames.ACCEPT).get.toLowerCase
```
derekspalmer, line 283: We shouldn't really use the .get but since we hit acceptHeaderActionProvider beforehand it will exist.

drnaylor-oc: Yeah... we have a couple of places like that -- but we always have it so `get` is actually okay here, I guess.

drnaylor-oc (dismissed):

drnaylor-oc (approved):

PR #473, opened 2023-09-21
[CTCP-3939] disable phase4 api calls when phase 5 goes to live

prashanthmamidi:


app/config/AppConfig.scala
```diff
@@ -62,4 +62,6 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: CTCServicesCon
 
   val internalAuthToken: String = config.get[String]("internal-auth.token")
 
+  val enablePhase5: Boolean = config.get[Boolean]("enable-phase-5")
```
drnaylor-oc, line 65: This is a bad config name, as it's not about enabling phase 5, it's about disabling phase 4.

app/routing/ArrivalsRouter.scala
```diff
@@ -60,7 +64,16 @@ class ArrivalsRouter @Inject() (
...
+      case _ =>
+        if (config.enablePhase5) handlePhase5()
+        else v1Arrivals.createArrivalNotification()
+    }
+
+  private def handlePhase5() =
+    Action(streamFromMemory) {
+      request =>
+        request.body.runWith(Sink.ignore)
+        Gone(Json.obj("message" -> "Please use version 2 to create Arrival Notification"))
```
drnaylor-oc, line 76: It would be far better to create a `gone` `PresentationError` and use that here, in the same style as we do in phase 5.

We also need to have better messages than this, it's worth asking Fergal what good wording might be.

app/routing/DeparturesRouter.scala
```diff
@@ -59,9 +63,18 @@ class DeparturesRouter @Inject() (
...
+    case _ =>
+      if (config.enablePhase5) handlePhase5()
+      else v1Departures.submitDeclaration()
   }
 
+  private def handlePhase5() =
+    Action(streamFromMemory) {
+      request =>
+        request.body.runWith(Sink.ignore)
+        Gone(Json.obj("message" -> "Please use version 2 to create Departure Declaration"))
```
drnaylor-oc, line 75: As above.

test/controllers/ArrivalMovementControllerSpec.scala
```diff
@@ -71,12 +74,15 @@ class ArrivalMovementControllerSpec
...
+
   override lazy val app = GuiceApplicationBuilder()
     .overrides(
       bind[Metrics].toInstance(new TestMetrics),
       bind[AuthAction].to[FakeAuthAction],
       bind[ArrivalConnector].toInstance(mockArrivalConnector),
-      bind[Clock].toInstance(mockClock)
+      bind[Clock].toInstance(mockClock),
+      bind[Configuration].toInstance(Configuration(configuration))
     )
```
drnaylor-oc, line 86: Isn't there a `configure` override we can just use instead?

Plus, why would we need to change it here, if the config entry isn't going to be read in this controller?

test/controllers/DeparturesControllerSpec.scala
```diff
@@ -88,7 +93,8 @@ class DeparturesControllerSpec
       bind[DeparturesConnector].toInstance(mockDepartureConnector),
       bind[EnsureGuaranteeService].toInstance(mockGuaranteeService),
       bind[AuditService].toInstance(mockAuditService),
-      bind[Clock].toInstance(mockClock)
+      bind[Clock].toInstance(mockClock),
+      bind[Configuration].toInstance(Configuration(configuration))
```
drnaylor-oc, line 97: As above, again, there shouldn't even be a need for this.

test/controllers/ArrivalMovementControllerSpec.scala
prashanthmamidi: the phase4 controller tests use the config from application.conf by default - It's all good if we have configured the property as `enable-phase-5 = false` , but when we change this to `true`, then  these phase4 controller tests will fail as we return `GONE` instead of `ACCEPTED`

Not aware of `configure` where we can override the properties -then we no need to create a separate config file

drnaylor-oc: Oh, of course, in the P4 tests it hits the endpoint as if it was a web server, rather than directly. Okay, that makes sense.

app/config/AppConfig.scala
prashanthmamidi: will update the name to `disable-phase-4`

app/routing/DeparturesRouter.scala
```diff
@@ -64,15 +65,16 @@ class DeparturesRouter @Inject() (
...
   }
 
-  private def handlePhase5() =
+  private def handleDisablingPhase4() =
     Action(streamFromMemory) {
       request =>
         request.body.runWith(Sink.ignore)
-        Gone(Json.obj("message" -> "Please use version 2 to create Departure Declaration"))
+        val presentationError = PresentationError.goneError("Please use CTC Traders API v2.0 to create a Departure Declaration")
+        Status(presentationError.code.statusCode)(Json.obj("message" -> presentationError.message))
```
drnaylor-oc, line 77: Status(presentationError.code.statusCode)(Json.obj(presentationError))

app/routing/ArrivalsRouter.scala
```diff
@@ -65,15 +66,16 @@ class ArrivalsRouter @Inject() (
...
     }
 
-  private def handlePhase5() =
+  private def handleDisablingPhase4() =
     Action(streamFromMemory) {
       request =>
         request.body.runWith(Sink.ignore)
-        Gone(Json.obj("message" -> "Please use version 2 to create Arrival Notification"))
+        val presentationError = PresentationError.goneError("Please use CTC Traders API v2.0 to create an Arrival Notification")
+        Status(presentationError.code.statusCode)(Json.obj("message" -> presentationError.message))
```
drnaylor-oc, line 78: Status(presentationError.code.statusCode)(Json.obj(presentationError))

drnaylor-oc (dismissed):
Before you merge this, please run the Phase 4 postman scripts against this in the default state, just to make sure we don't impact phase 4 before we intend to.

sonisn1988 (approved):

PR #474, opened 2023-10-04
Create AuditService method to audit message type audit messages (comm…

sonisn1988:
…on-transit-convention-traders)

### DIFF
--- a/app/config/Constants.scala
+++ b/app/config/Constants.scala
@@ -35,6 +35,18 @@ object Constants {
 
   val XContentLengthHeader = "X-ContentLength"
 
+  val XMovementIdHeader = "X-Audit-Meta-Movement-Id"
+
+  val XMessageIdHeader = "X-Audit-Meta-Message-Id"
+
+  val XEoriHeader = "X-Audit-Meta-EORI"
+
+  val XMovementTypeHeader = "X-Audit-Meta-Movement-Type"
+
+  val XURIPathHeader = "X-Audit-Meta-Path"
+
+  val XAuditSourceHeader = "X-Audit-Source"
+
   val ChannelHeader = "channel"
 
   val LegacyEnrolmentKey: String   = "HMCE-NCTS-ORG"
--- a/app/config/Constants.scala
+++ b/app/config/Constants.scala
@@ -35,18 +35,6 @@ object Constants {
 
   val XContentLengthHeader = "X-ContentLength"
 
-  val XMovementIdHeader = "X-Audit-Meta-Movement-Id"
-
-  val XMessageIdHeader = "X-Audit-Meta-Message-Id"
-
-  val XEoriHeader = "X-Audit-Meta-EORI"
-
-  val XMovementTypeHeader = "X-Audit-Meta-Movement-Type"
-
-  val XURIPathHeader = "X-Audit-Meta-Path"
-
-  val XAuditSourceHeader = "X-Audit-Source"
-
   val ChannelHeader = "channel"
 
   val LegacyEnrolmentKey: String   = "HMCE-NCTS-ORG"
--- a/app/v2/connectors/AuditingConnector.scala
+++ b/app/v2/connectors/AuditingConnector.scala
@@ -116,16 +116,16 @@ class AuditingConnectorImpl @Inject() (httpClient: HttpClientV2, val metrics: Me
       httpClient
         .post(url"$url")
         .withInternalAuthToken
+        .withMovementId(movementId)
+        .withMessageId(messageId)
+        .withEoriNumber(enrolmentEORI)
+        .withMovementType(movementType)
+        .withMessageType(messageType)
         .setHeader(
           HeaderNames.CONTENT_TYPE       -> contentType,
           Constants.XContentLengthHeader -> contentLength.toString,
-          Constants.XMovementIdHeader    -> movementId.map(_.value).getOrElse(None).toString,
-          Constants.XMessageIdHeader     -> messageId.map(_.value).getOrElse(None).toString,
-          Constants.XEoriHeader          -> enrolmentEORI.map(_.value).getOrElse(None).toString,
-          Constants.XMovementTypeHeader  -> movementType.map(_.movementType).getOrElse(None).toString,
-          Constants.XMessageTypeHeader   -> messageType.map(_.code).getOrElse(None).toString,
-          Constants.XURIPathHeader       -> path,
-          Constants.XAuditSourceHeader   -> "common-transit-convention-traders"
+          "X-Audit-Meta-Path"            -> path,
+          "X-Audit-Source"               -> "common-transit-convention-traders"
         )
         .withBody(payload)
         .executeAndExpect(ACCEPTED)
--- a/app/v2/connectors/AuditingConnector.scala
+++ b/app/v2/connectors/AuditingConnector.scala
@@ -31,7 +31,12 @@ import play.api.http.Status.ACCEPTED
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.http.StringContextOps
 import uk.gov.hmrc.http.client.HttpClientV2
+import v2.models.request.MessageType
 import v2.models.AuditType
+import v2.models.EORINumber
+import v2.models.MessageId
+import v2.models.MovementId
+import v2.models.MovementType
 
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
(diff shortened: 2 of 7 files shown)


app/v2/connectors/AuditingConnector.scala
```diff
@@ -71,4 +92,43 @@ class AuditingConnectorImpl @Inject() (httpClient: HttpClientV2, val metrics: Me
...
+          case ("path", value) => value
+        }
+        .getOrElse("-")
+      httpClient
+        .post(url"$url")
+        .withInternalAuthToken
+        .setHeader(
+          HeaderNames.CONTENT_TYPE       -> contentType,
+          Constants.XContentLengthHeader -> contentLength.toString,
+          Constants.XMovementIdHeader    -> movementId.map(_.value).getOrElse(None).toString,
```
drnaylor-oc, line 122:
```scala
@ None.toString
res1: String = "None"
```

This isn't what we want. If an element doesn't exist, the header shouldn't exist.

drnaylor-oc: Also, that `getOrElse(None)` is pointless. If the original option is `None`, `option.map(...)` is still `None`.

it/v2/connectors/AuditingConnectorSpec.scala
```diff
@@ -135,4 +144,116 @@ class AuditingConnectorSpec
...
+                    )
+                      .withHeader(HeaderNames.AUTHORIZATION, equalTo(token))
+                      .withHeader(HeaderNames.CONTENT_TYPE, equalTo(contentType))
+                      .withHeader(Constants.XContentLengthHeader, equalTo(contentSize.toString))
+                      .withHeader(Constants.XMovementIdHeader, equalTo(movementId.map(_.value).getOrElse(None).toString))
+                      .withHeader(Constants.XMessageIdHeader, equalTo(messageId.map(_.value).getOrElse(None).toString))
+                      .withHeader(Constants.XEoriHeader, equalTo(eori.map(_.value).getOrElse(None).toString))
+                      .withHeader(Constants.XURIPathHeader, equalTo("/customs/transits/movements"))
+                      .withHeader(Constants.XMessageTypeHeader, equalTo(messageType.map(_.code).getOrElse(None).toString))
+                      .withHeader(Constants.XMovementTypeHeader, equalTo(movementType.map(_.movementType).getOrElse(None).toString))
```
drnaylor-oc, line 218: This is something that, as a team, we've been quite bad at in testing -- and wouldn't be caught here.

The problem we have here is that you've just replicated the line above and what it didn't catch is that the header value would be `None`, which is not useful. I suggest making a couple of tests, one that has the minimal set of headers, the maximum set of headers, a selection of headers, and then a test like the above. 

I get it's tough to test this with property based testing and to test it in a different way, but we do need to make sure it works as intended without just duplicating the actual code.

test/v2/services/AuditingServiceSpec.scala
```diff
@@ -79,4 +95,123 @@ class AuditingServiceSpec extends AnyFreeSpec with Matchers with ScalaFutures wi
...
+  "For auditing message and status type event" - {
+    "Posting any audit type event" - Seq(MimeTypes.XML, MimeTypes.JSON).foreach {
+      contentType =>
+        s"when contentType equals $contentType" - {
+          "on success audit, return the successful future" in forAll(
+            Gen.option(arbitrary[EORINumber]),
+            Gen.option(arbitrary[MovementType]),
+            Gen.option(arbitrary[MovementId]),
+            Gen.option(arbitrary[MessageId]),
+            Gen.option(arbitrary[MessageType])
```
drnaylor-oc, line 107: I _do_ appreciate the use of `Gen.option` here -- I'm pleased we're doing that.

app/config/Constants.scala
```diff
@@ -35,6 +35,18 @@ object Constants {
...
+  val XMessageIdHeader = "X-Audit-Meta-Message-Id"
+
+  val XEoriHeader = "X-Audit-Meta-EORI"
+
+  val XMovementTypeHeader = "X-Audit-Meta-Movement-Type"
+
+  val XURIPathHeader = "X-Audit-Meta-Path"
+
+  val XAuditSourceHeader = "X-Audit-Source"
+
```
drnaylor-oc, line 49: These constants are only appropriate in the audit connector, and not used elsewhere in the service. I feel like they should be in there instead, rather than globally available.

app/v2/connectors/V2BaseConnector.scala
```diff
@@ -145,6 +140,31 @@ trait V2BaseConnector extends HttpErrorFunctions {
...
+
+    def withMessageType(messageType: Option[MessageType]): RequestBuilder =
+      if (messageType.isDefined)
+        requestBuilder.setHeader("X-Audit-Meta-Message-Type" -> messageType.get.code)
+      else requestBuilder
+
+    def withMessageId(messageId: Option[MessageId]): RequestBuilder =
+      if (messageId.isDefined)
+        requestBuilder.setHeader("X-Audit-Meta-Message-Id" -> messageId.get.value)
+      else requestBuilder
```
drnaylor-oc, line 166: Now you've done this (and I approve!) we can test these in isolation too -- let's make a set of unit tests for these alone.

test/v2/connectors/V2BaseConnectorSpec.scala
```diff
@@ -300,6 +300,102 @@ class V2BaseConnectorSpec
...
+        verify(sut, times(1)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-Movement-Id" -> movementId.value))): _*)
+    }
+
+    "withMovementId ignore the audit movement Id header for None value" in {
+      val sut = mock[RequestBuilder]
+      // any here, verify later
+      when(sut.setHeader(any())).thenReturn(sut)
+
+      sut.withMovementId(None)
+      verify(sut, times(0)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-Movement-Id" -> "None"))): _*)
```
drnaylor-oc, line 319: That's not quite right -- we want to verify it's not called at all, no matter the arguments. All you're verifying here is that this isn't called with this specific argument list, which is true, but

```scala
        verify(sut, times(1)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-Movement-Id" -> movementId.value))): _*)
```

could still be true.

Something like

```scala
        verify(sut, times(0)).setHeader(ArgumentMatchers.eq(any()))
```

is what you really want.

test/v2/connectors/V2BaseConnectorSpec.scala
```diff
@@ -300,6 +300,102 @@ class V2BaseConnectorSpec
...
+        verify(sut, times(1)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-Movement-Type" -> movementType.movementType))): _*)
+    }
+
+    "withMovementType ignore the audit movement type header for None value" in {
+      val sut = mock[RequestBuilder]
+      // any here, verify later
+      when(sut.setHeader(any())).thenReturn(sut)
+
+      sut.withMovementType(None)
+      verify(sut, times(0)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-Movement-Type" -> "None"))): _*)
```
drnaylor-oc, line 338: and here

test/v2/connectors/V2BaseConnectorSpec.scala
```diff
@@ -300,6 +300,102 @@ class V2BaseConnectorSpec
...
+        verify(sut, times(1)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-EORI" -> eoriNumber.value))): _*)
+    }
+
+    "withEoriNumber ignore the audit Eori number header for None value" in {
+      val sut = mock[RequestBuilder]
+      // any here, verify later
+      when(sut.setHeader(any())).thenReturn(sut)
+
+      sut.withEoriNumber(None)
+      verify(sut, times(0)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-EORI" -> "None"))): _*)
```
drnaylor-oc, line 357: and here

test/v2/connectors/V2BaseConnectorSpec.scala
```diff
@@ -300,6 +300,102 @@ class V2BaseConnectorSpec
...
+        verify(sut, times(1)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-Message-Id" -> messageId.value))): _*)
+    }
+
+    "withMessageId ignore the audit message Id header None value" in {
+      val sut = mock[RequestBuilder]
+      // any here, verify later
+      when(sut.setHeader(any())).thenReturn(sut)
+
+      sut.withMessageId(None)
+      verify(sut, times(0)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-Message-Id" -> "None"))): _*)
```
drnaylor-oc, line 376: and here

test/v2/connectors/V2BaseConnectorSpec.scala
```diff
@@ -300,6 +300,102 @@ class V2BaseConnectorSpec
...
+    }
+
+    "withMessageType ignore the audit message type header for None value" in forAll(arbitrary[MessageType]) {
+      messageType =>
+        val sut = mock[RequestBuilder]
+        // any here, verify later
+        when(sut.setHeader(any())).thenReturn(sut)
+
+        sut.withMessageType(None)
+        verify(sut, times(0)).setHeader(ArgumentMatchers.eq(Seq(("X-Audit-Meta-Message-Type" -> "None"))): _*)
```
drnaylor-oc, line 396: and here

test/v2/connectors/V2BaseConnectorSpec.scala
drnaylor-oc: It's not true here specifically, but It _could_ be true if you introduced that as a bug in the code.

I guess the problem is that it's a variardic parameter. I don't know how to work around it, but I don't have the answers to everything. I just want the tests to be properly representative of what happens.

You might be able to just use `verifyNoInteractions` (or whatever it's called) instead.

drnaylor-oc (approved):
Thanks for working with me in getting this into a better shape. It looks better now. I don't think there are any issues here, but if there are we'll likely pick them up once we hook up the new events.

PR #478, opened 2023-10-12
CTCP-3737: Remove documentation regarding duplicate LRNs

drnaylor-oc:
Also update definition regarding the 8MB uplift.

### DIFF
--- a/resources/public/api/conf/2.0/application.yaml
+++ b/resources/public/api/conf/2.0/application.yaml
@@ -9,7 +9,7 @@ info:
     and to get messages sent from customs offices of departure and destination.
     <br/><br/>
     You can use this version of the API to send both small messages (up to 5MB in size) and large messages 
-    (up to 5MB in size - this limit will later increase to 8MB). The large messages capability applies only to POST endpoints. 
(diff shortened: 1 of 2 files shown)


resources/public/api/conf/2.0/application.yaml
```diff
@@ -1388,21 +1374,6 @@ components:
...
-        - message
-      type: object
-      properties:
-        code:
-          enum:
-            - CONFLICT
-          type: string
-        message:
-          type: string
-          description: The description of the error.
```
fmcgrath, line 1405: Did we miss this for 11 October comms and content?

fmcgrath (approved):
Looks OK.

resources/public/api/conf/2.0/application.yaml
drnaylor-oc: No, this bit is just the LRN reuse stuff so it's more supplementary to it - I mentioned that the 409 is no longer sent on a duplicate LRN, but an IE056 is. I just forgot to change it here.

fmcgrath: OK.

PR #479, opened 2023-10-16
CTCP-4180 Create audit event for success event (common-transit-convention-traders) and CTCP-4181 Extra P4 audit events for failed scenario(api change)

sonisn1988:


test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -406,7 +407,17 @@ class V2MovementsControllerSpec
             .thenAnswer {
               _ => EitherT.rightT(())
             }
-
+          when(
```
drnaylor-oc, line 410: Can we also add tests to verify that in some error scenarios, this is NOT called?

derekspalmer (approved):

PR #480, opened 2023-10-18
[CTCP-3543] added audit events for failed scenarios

prashanthmamidi:


app/v2/controllers/V2MovementsController.scala
```diff
@@ -172,8 +178,20 @@ class V2MovementsControllerImpl @Inject() (
...
         implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
         (for {
-          _               <- contentSizeIsLessThanLimit(size)
-          _               <- validationService.validateXml(MessageType.DeclarationData, request.body).asPresentation
+          _ <- contentSizeIsLessThanLimit(size)
+          _ <- validationService.validateXml(MessageType.DeclarationData, request.body).asPresentation.leftMap {
+            err =>
+              auditService.auditStatusEvent(
+                ValidationFailed,
+                Some(Json.obj("message" -> err.message)),
```
sonisn1988, line 186: Here we need to pass ```Some(Json.toJson(err))```

app/controllers/PushPullNotificationController.scala
```diff
@@ -45,8 +52,13 @@ class PushPullNotificationController @Inject() (cc: ControllerComponents, pushPu
         case Some(clientId) =>
           pushPullNotificationConnector.getBox(clientId).map {
             case Left(error) if error.statusCode == NOT_FOUND =>
-              NotFound(Json.toJson(JsonClientErrorResponse(NOT_FOUND, "No box found for your client id")))
-            case Left(_)    => InternalServerError(Json.toJson(JsonSystemErrorResponse(INTERNAL_SERVER_ERROR, "Unexpected Error")))
+              val errorMessage = Json.toJson(JsonClientErrorResponse(NOT_FOUND, "No box found for your client id"))
```
sonisn1988, line 55: This one you have done in v1 and that is not part of this ticket I believe.

prashanthmamidi: it's one of the failed audit defined in the confluence `https://confluence.tools.tax.service.gov.uk/pages/viewpage.action?spaceKey=CTCTA&title=Auditing%3A+Architecture+and+Structure`

drnaylor-oc: Sandeep is right though -- we shouldn't be touching v1 in this ticket. All of the auditing we need to do is in v2.

sonisn1988: Yes I knew that. I think you only had created this list of event with audit spike. As this is for getBox and that part we are doing in ```transit-movements-push-notifications``` for P5. so if you really want to do that then this should go inside ```transit-movements-push-notifications```

drnaylor-oc: That's not so true. We can still do that in `common-transit-convention-traders` in response to the call to get the box.

prashanthmamidi: ok, will remove this event from v1 code

app/v2/controllers/V2MovementsController.scala
```diff
@@ -174,8 +186,20 @@ class V2MovementsControllerImpl @Inject() (
       implicit request => size =>
         implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
         (for {
-          _               <- contentSizeIsLessThanLimit(size)
-          _               <- validationService.validateXml(MessageType.DeclarationData, request.body).asPresentation
+          _ <- contentSizeIsLessThanLimit(size)
+          _ <- validationService.validateXml(MessageType.DeclarationData, request.body).asPresentation.leftMap {
```
sonisn1988, line 190: I think we also need to add ```ValidationFailed``` for validateJson method?

drnaylor-oc: Yes, we do.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -243,13 +279,37 @@ class V2MovementsControllerImpl @Inject() (
...
             .upscanInitiate(request.eoriNumber, movementType, movementResponse.movementId, movementResponse.messageId)
             .asPresentation
           boxResponseOption <- mapToOptionalResponse(
             pushNotificationsService.associate(movementResponse.movementId, movementType, request.headers, request.eoriNumber)
-          )
+          ).leftMap {
+            err =>
+              auditService.auditStatusEvent(
+                PushNotificationFailed,
+                None,
```
sonisn1988, line 304: do we not need to pass ```Some(Json.toJson(err))``` here as payload?

app/v2/controllers/V2MovementsController.scala
```diff
@@ -579,9 +702,21 @@ class V2MovementsControllerImpl @Inject() (
...
-          messageType            <- xmlParsingService.extractMessageType(request.body, messageTypeList).asPresentation
-          _                      <- validationService.validateXml(messageType, request.body).asPresentation
+          _           <- contentSizeIsLessThanLimit(size)
+          messageType <- xmlParsingService.extractMessageType(request.body, messageTypeList).asPresentation
+          _ <- validationService.validateXml(messageType, request.body).asPresentation.leftMap {
+            err =>
+              auditService.auditStatusEvent(
+                ValidationFailed,
+                Some(Json.toJson(err)),
+                None,
```
sonisn1988, line 712: Please add the ```movementId``` instead of ```None``` here

app/v2/controllers/V2MovementsController.scala
```diff
@@ -869,7 +1086,19 @@ class V2MovementsControllerImpl @Inject() (
         Some(messageType)
       )
       boxResponseOption <- mapToOptionalResponse[PushNotificationError, BoxResponse](
-        pushNotificationsService.associate(movementResponse.movementId, movementType, request.headers, request.eoriNumber)
+        pushNotificationsService.associate(movementResponse.movementId, movementType, request.headers, request.eoriNumber).leftMap {
+          err =>
+            auditService.auditStatusEvent(
+              PushNotificationFailed,
+              None,
```
sonisn1988, line 1093: Need to pass the error payload not ```None```

app/v2/controllers/V2MovementsController.scala
```diff
@@ -796,6 +979,16 @@ class V2MovementsControllerImpl @Inject() (
               updateMovementResponse.messageId,
               MessageStatus.Failed
             )
+            auditService.auditStatusEvent(
+              SubmitAttachMessageFailed,
+              Some(Json.toJson(err)),
+              Some(movementId),
+              None,
```
sonisn1988, line 986: pass ```updateMovementResponse.messageId``` not ```None``` here

app/controllers/PushPullNotificationController.scala
sonisn1988: can not see ```PushPullNotificationGetBoxFailed``` with version 2 so still this is not resolved looks like.

prashanthmamidi: we get the `BoxResponse[BoxId]` through `postAssociation()`  call for which we already have failed event as `PushNotificationFailed`

sonisn1988: @drnaylor-oc are you suggesting to implement this ```PushPullNotificationGetBoxFailed``` for ```NOT_FOUND``` case in ```common-transit-convention-traders``` for ```associate()``` method PPNS service call?

app/v2/controllers/V2MovementsController.scala
```diff
@@ -243,12 +303,38 @@ class V2MovementsControllerImpl @Inject() (
...
             .asPresentation
           boxResponseOption <- mapToOptionalResponse(
-            pushNotificationsService.associate(movementResponse.movementId, movementType, request.headers, request.eoriNumber)
+            pushNotificationsService
+              .associate(movementResponse.movementId, movementType, request.headers, request.eoriNumber)
+              .leftMap {
+                err =>
+                  auditService.auditStatusEvent(
+                    PushNotificationFailed,
+                    Some(Json.obj("message" -> err.toString)),
```
sonisn1988, line 329: here we need to pass ```Json.toJson(err)```

prashanthmamidi: This is added on purpose as `err` is not presentation an error - it's a `PushNotificationError`  which is an obj on it's own without any params. To get a proper json we have go with this approach.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -783,7 +992,19 @@ class V2MovementsControllerImpl @Inject() (
...
+      _ = pushNotificationsService.update(movementId).leftMap {
+        err =>
+          auditService.auditStatusEvent(
+            PushNotificationUpdateFailed,
+            None,
+            Some(movementId),
+            Some(updateMovementResponse.messageId),
+            Some(request.eoriNumber),
+            Some(movementType),
+            None
```
sonisn1988, line 1004: At this point of time we will be having messageType available so please pass that.

app/controllers/PushPullNotificationController.scala
drnaylor-oc: I just did a search of the code in my IDE:

* `PushNotificationFailed` is defined, but not used anywhere. According to your spikes and what I pulled together at https://confluence.tools.tax.service.gov.uk/display/CTCTA/Auditing%3A+Architecture+and+Structure, this should be when an actual push notification fails. This will mostly happen from the router, but _can_ happen from here too.
* `PushPullNotificationGetBoxFailed` is for when the association fails, so as @sonisn1988 said.

drnaylor-oc:
This is missing unit tests verifying that said audit events are being created, please add those too.

app/v2/controllers/V2MovementsController.scala
sonisn1988: Ohk my bad. @drnaylor-oc is this ok for you? or need to change it to ```None```?

app/controllers/PushPullNotificationController.scala
prashanthmamidi: `PushNotificationFailed` is defined maybe update the code  - this is added for when association fails.
Will use `PushPullNotificationGetBoxFailed` instead of `PushNotificationFailed` for when the association fails.

So, `PushNotificationFailed` in CTC traders where we should use this?

app/v2/controllers/V2MovementsController.scala
drnaylor-oc: Oh, I see, it's used here. My search skills are bad these days.

Whatever makes the most sense here.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -889,6 +1158,15 @@ class V2MovementsControllerImpl @Inject() (
               movementResponse.messageId,
               MessageStatus.Failed
             )
+            auditService.auditStatusEvent(
+              if (movementType.movementType == "departure") SubmitDeclarationFailed else SubmitArrivalNotificationFailed,
```
sonisn1988, line 1162: Please use ```movementType == MovementType.Departure``` as suggested earlier.

app/controllers/PushPullNotificationController.scala
sonisn1988: ```PushNotificationFailed``` you are going to use for ```InternalServerError``` and ```PushPullNotificationGetBoxFailed``` you are going to use for ```NotFound``` case for the same PPNS call.

prashanthmamidi: As mentioned earlier,  the err here is `PushNotificationError` not PresentationError, so we can't check the status code here. 
And also don't like the idea of this as well - if we do separate cases for each and every error code then we have to do this for the rest of the project too. We should just do one generic failed event.

On the other hand, the original intention to add `PushPullNotificationGetBoxFailed` in the spike is for v1 endpoint  as this is not applicable for us.

So,  for `associate()` failed scenario- either we go with PushPullNotificationGetBoxFailed or PushNotificationFailed or some other proper audit event name which is more generic.

test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -978,6 +1020,17 @@ class V2MovementsControllerSpec
             eqTo(Some(MovementType.Departure)),
             eqTo(Some(MessageType.DeclarationData))
           )(any[HeaderCarrier], any[ExecutionContext])
+
+          verify(mockAuditService, times(1)).auditStatusEvent(
+            eqTo(SubmitDeclarationFailed),
+            eqTo(Some(Json.obj("message" -> "Internal server error", "code" -> "INTERNAL_SERVER_ERROR"))),
```
sonisn1988, line 1026: Same here no need to construct json object, just use ```Json.toJson(PresentationError.internalServiceError())```

test/v2/controllers/V2MovementsControllerSpec.scala
```diff
@@ -843,6 +875,16 @@ class V2MovementsControllerSpec
           "code"    -> "INTERNAL_SERVER_ERROR",
           "message" -> "Internal server error"
         )
+
+        verify(mockAuditService, times(1)).auditStatusEvent(
+          eqTo(CreateMovementDBFailed),
+          eqTo(Some(Json.obj("message" -> "Internal server error", "code" -> "INTERNAL_SERVER_ERROR"))),
```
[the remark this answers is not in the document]
prashanthmamidi: this is a test and I prefer to be as is as this is much readable

test/v2/controllers/V2MovementsControllerSpec.scala
prashanthmamidi: same

app/controllers/PushPullNotificationController.scala
sonisn1988: You will be receiving error type like ```PushNotificationError.BoxNotFound``` and ```PushNotificationError.UnexpectedError``` and then take the decision.  Probably Dan can suggest if we don't want to implement ```PushPullNotificationGetBoxFailed``` here or can suggest.

test/v2/controllers/V2MovementsControllerSpec.scala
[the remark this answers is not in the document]
sonisn1988: Ok no problem as your wish in that case.

app/controllers/PushPullNotificationController.scala
drnaylor-oc: I'd rather see it implemented as was spiked. We agreed on these audit types, let's try to implement them as suggested in the first place.

app/v2/controllers/V2MovementsController.scala
```diff
@@ -243,12 +304,38 @@ class V2MovementsControllerImpl @Inject() (
...
+                None
+              )
+              err
+          }
           upscanResponse <- upscanService
             .upscanInitiate(request.eoriNumber, movementType, movementResponse.movementId, movementResponse.messageId)
             .asPresentation
           boxResponseOption <- mapToOptionalResponse(
-            pushNotificationsService.associate(movementResponse.movementId, movementType, request.headers, request.eoriNumber)
+            pushNotificationsService
```
sonisn1988, line 324: Please add  ```val auditType = if (err == PushNotificationError.BoxNotFound) PushPullNotificationGetBoxFailed else PushNotificationFailed``` here as well.

sonisn1988 (approved):

PR #482, opened 2023-10-27
[CTCP-4215] config gate to stop PD alerts

prashanthmamidi:


app/v2/controllers/V2MovementsController.scala
```diff
@@ -821,139 +829,147 @@ class V2MovementsControllerImpl @Inject() (
...
       implicit request =>
-        val originalHc: HeaderCarrier = HeaderCarrierConverter.fromRequest(request)
-
-        // If the client ID is provided, ensure we add it to the headers.
-        val clientIdHeader = clientId
-          .map[Seq[(String, String)]](
-            id => Seq(XClientIdHeader -> id.value)
-          )
-          .getOrElse(Seq.empty)
+        if (config.enablePhase5) {
```
drnaylor-oc, line 832: No point doing it here as this is only from Upscan, which will never be triggered.

test/routing/ArrivalsRouterSpec.scala
```diff
@@ -84,27 +85,83 @@ class ArrivalsRouterSpec extends AnyFreeSpec with Matchers with OptionValues wit
...
-            contentAsJson(result) mustBe Json.obj("version" -> 2) // ensure we get the unique value to verify we called the fake action
+
+            isVersion match {
+              case true => contentAsJson(result) mustBe Json.obj("version" -> 2) // ensure we get the unique value to verify we called the fake action
+              case false =>
+                contentAsJson(result) mustBe Json.obj(
+                  "message" -> "CTC Traders API version 2 is not yet available. Please continue to use version 1 to submit transit messages.",
+                  "code"    -> "NOT_ACCEPTABLE"
+                )
+            }
```
drnaylor-oc, line 106: Can we just use an `if` statement here? We really shouldn't use pattern matching on simple booleans...

drnaylor-oc (approved):

PR #489, opened 2024-01-11
[CTCP-4701] Reinstated logging in application-json-logger.xml

brianbayes:


conf/application-json-logger.xml
```diff
@@ -5,4 +5,8 @@
         <encoder class="uk.gov.hmrc.play.logging.JsonEncoder"/>
     </appender>
```
sonisn1988, line 7: You need to include logger package name as well I think ?
something like ``` <logger name="uk.gov" level="WARN"/>```. Its looks like you only define the log level.

derekspalmer (approved):

PR #493, opened 2024-05-28
Create helpful message rather than - “Uploaded file not accepted” to ……ppns

sonisn1988:


derekspalmer (approved):
I would have possibly put:
badRequestError(s"Uploaded file not accepted: ${failureDetails.message}")

PR #497, opened 2024-07-16
Fix compile warnings

DamienButterworth:
This PR is to fix compile warnings in the service


app/config/AppConfig.scala
```diff
@@ -30,25 +30,25 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: CTCServicesCon
   val auditingEnabled: Boolean = config.get[Boolean]("auditing.enabled")
   val graphiteHost: String     = config.get[String]("microservice.metrics.graphite.host")
```
sonisn1988, line 31: I think we can remove unused reference as well from this file. That I think should be reported as part of compile warnings?

DamienButterworth: I agree we should remove unused references. However, wasn't being reported as a compile warnings interestingly, will have a look for unused references across the codebase

sonisn1988: Also please have a look on ```Constants.scala ```. Looks like there also we need to add type reference and need to remove some unused references.

app/v2/controllers/ConvertError.scala
```diff
@@ -51,84 +64,52 @@ trait ConvertError extends Logging {
     def convert(input: E): PresentationError
   }
 
-  val jsonToXmlValidationErrorConverter = new Converter[FailedToValidateError] {
-    import v2.models.errors.FailedToValidateError._
```
sonisn1988, line 55: I think we can keep import at block level if it's not reported in warnings. That's look simple and more understandable.

DamienButterworth: This has already been changed 👍

app/v2/controllers/V2MovementsController.scala
```diff
@@ -440,20 +444,30 @@ class V2MovementsControllerImpl @Inject() (
...
+            case VERSION_2_ACCEPT_HEADER_VALUE_JSON =>
+              for {
+                jsonStream <- conversionService.xmlToJson(messageSummary.messageType.get, bodyAndSize.body).asPresentation
+                bodyWithContentType = BodyAndContentType(MimeTypes.JSON, jsonStream)
+              } yield bodyWithContentType
+            case VERSION_2_ACCEPT_HEADER_VALUE_XML =>
+              EitherT.rightT[Future, PresentationError](
+                BodyAndContentType(MimeTypes.XML, bodyAndSize.body)
+              )
+            //TODO: These two cases were not caught previously should we be handling these?
```
sonisn1988, line 463: I don't think so it's required as if you look on entry point for this method call i.e. ```getMessageBody``` we already have ```acceptHeaderActionProvider``` that handle all case condition.  Also I think we need to use ```VERSION_2_ACCEPT_HEADER_VALUE_JSON.value```  for above case conditions?

app/v2/controllers/V2MovementsController.scala
```diff
@@ -472,30 +486,38 @@ class V2MovementsControllerImpl @Inject() (
...
+                EitherT.rightT[Future, PresentationError](
+                  Source.single(ByteString(Json.stringify(HateoasMovementMessageResponse(movementId, messageSummary.id, messageSummary, movementType))))
+                )
+              else
+                mergeStreamIntoJson(
+                  Json.toJson(HateoasMovementMessageResponse(movementId, messageSummary.id, messageSummary, movementType)).as[JsObject].fields,
+                  "body",
+                  bodyAndSize.body
+                )
+            case VERSION_2_ACCEPT_HEADER_VALUE_XML =>
```
sonisn1988, line 516: We don't need this. Please see the above comments

DamienButterworth: We may not need it in terms of functionality. However, it makes the match case non-exhaustive if we don't handle it I could change it to be a catch-all instead if you prefer? just to make the match statement exhaustive

app/v2/controllers/V2MovementsController.scala
DamienButterworth: We may not need to handle it in terms of functionality. However, the match statement should be exhaustive, We should require it to have `.value` because the `formatAccept` now returns the models instead of a string

app/v2/controllers/V2MovementsController.scala
sonisn1988: Yes sure. Thank you :)

app/v2/models/MessageStatus.scala
```diff
@@ -26,7 +26,7 @@ import play.api.libs.json.Writes
 sealed trait MessageStatus
 
 object MessageStatus {
-  final case object Received extends MessageStatus
+  final private case object Received extends MessageStatus
```
sonisn1988, line 29: do we need to make this as ```private```?

DamienButterworth: We don't strictly need to, it just wasn't being used elsewhere, i'll change it back

app/utils/Utils.scala
```diff
@@ -17,13 +17,10 @@
...
 import java.net.URI
-import java.net.URLEncoder
 
 object Utils {
 
   def lastFragment(location: String): String =
     URI.create(location).getPath.split("/").last
 
-  def urlEncode(str: String): String =
-    URLEncoder.encode(str, "UTF-8")
```
sonisn1988, line 28: is this not required?

sonisn1988 (approved):

app/utils/Utils.scala
DamienButterworth: It wasn't used by anything so I'm assuming not

PR #500, opened 2024-07-30
CTCP-5741: Routing based on headers

DamienButterworth:


sonisn1988 (approved):

app/v2_1/connectors/PersistenceConnector.scala
```diff
@@ -0,0 +1,380 @@
...
+  override def getMessage(eori: EORINumber, movementType: MovementType, movementId: MovementId, messageId: MessageId)(implicit
+    hc: HeaderCarrier,
+    ec: ExecutionContext
+  ): Future[MessageSummary] = {
+    val url = appConfig.movementsUrl.withPath(getMessageUrl(eori, movementType, movementId, messageId))
+    httpClientV2
+      .get(url"$url")
+      .withInternalAuthToken
+      .setHeader(HeaderNames.CONTENT_TYPE -> MimeTypes.XML)
+      .setHeader(Constants.APIVersionHeaderKey -> Constants.APIVersionFinalHeaderValue)
```
sonisn1988, line 180: as we are not using this additional header in transit-movements so this file can be removed. Please use the v2 file for this connector and service.

DamienButterworth: I propose we leave this in for the time being until we can verify we definitely don't require to identify transitional/final in the database, sending the header and not processing it won't cause any issues

PR #508, opened 2024-08-29
Added conditional logic around which API definition is used

DamienButterworth:


app/controllers/documentation/DocumentationController.scala
```diff
@@ -16,18 +16,28 @@
...
+  def definition(): Action[AnyContent] = {
+    if (appConfig.version21BetaEnabled) {
+      assets.at("/public/api", "v2_1-definition.json")
+    } else { assets.at("/public/api", "definition.json") }
+  }
 
-  def raml(version: String, file: String): Action[AnyContent] =
-    assets.at(s"/public/api/conf/$version", file)
+  def raml(version: String, file: String): Action[AnyContent] = {
+    if (!appConfig.version21BetaEnabled && version == "2.1") {
```
prashanthmamidi, line 39: is the condition should _appConfig.version21BetaEnabled_  like other method instead _!appConfig.version21BetaEnabled_

DamienButterworth: The reason this is done this way is because there's only one scenario where we should return NOT_FOUND and that's when the beta isn't enabled and the version is 2.1 if I were to change it to appConfig.Version21BetaEnabled and version == "2.1" then there'll be duplicate code e.g.

DamienButterworth: `(version, appConfig.version21BetaEnabled) match {
      case ("2.1", false) => Action.async(Future.successful(NotFound("Resource not found by Assets controller")))
      case ("2.1", true) =>  assets.at(s"/public/api/conf/$version", file)
      case ("2.0", _) => assets.at(s"/public/api/conf/$version", file)
    }`

prashanthmamidi (approved):

PR #547, opened 2026-07-08
IE022 - API development (linked to AEST1-510, AEST1-511)

sandeep-k-soni:


app/controllers/MovementsController.scala
```diff
@@ -518,14 +518,31 @@ class MovementsControllerImpl @Inject() (
...
+    messageSummary: MessageSummary
+  )(implicit
+    hc: HeaderCarrier
+  ) =
+    if (
+      movementType == MovementType.Departure && messageSummary.messageType.isDefined && MessageType.NotificationToAmendDeclaration == messageSummary.messageType.get && !config.enableMessageType022
+    )
+      EitherT.leftT[Future, Option[BodyAndSize]](PresentationError.internalServiceError("IE022 MessageType not supported"))
+    else getBody(request.authenticatedRequest.eoriNumber, movementType, movementId, messageId, messageSummary.body, version)
+
```
KiraNamida, line 545: Do we have a test plan for this as not seeing changes within the solution to account for this?

sandeep-k-soni: I had tried to add the test cases for this feature toggle. But the test for this class is flaky and very huge and result was inconsistent after adding the new test cases. So thought of testing and verifying this manually and via regression test repo. Also this is temporary things and would not impact much as we will be removing this once we go live. Please let me know if you are ok or need to put effort to create test cases for this.

app/controllers/MovementsController.scala
```diff
@@ -518,14 +518,31 @@ class MovementsControllerImpl @Inject() (
...
+    movementId: MovementId,
+    messageId: MessageId,
+    request: ValidatedVersionedRequest[AnyContent],
+    version: Version,
+    messageSummary: MessageSummary
+  )(implicit
+    hc: HeaderCarrier
+  ) =
+    if (
+      movementType == MovementType.Departure && messageSummary.messageType.isDefined && MessageType.NotificationToAmendDeclaration == messageSummary.messageType.get && !config.enableMessageType022
```
prashanthmamidi, line 541: add the test case for this

app/controllers/MovementsController.scala
sandeep-k-soni: @KiraNamida Added Managed to pass the tests.

app/controllers/MovementsController.scala
sandeep-k-soni: Added test cases.

KiraNamida (approved):
With the decision to remain on version 3.0 to support SWDs then we can now progress with the feature switch for our own benefit to decouple deployments from the new features needing to be complete.
:shipit:

